PR #9, opened 2019-06-18
Fix example to work again after refactors

trishume:
Various refactors broke the simple example, this fixes it.


makepaddev:
I have it all fixed up and updated locally, but wasm32 build is also broken for a bit. I'm working through a big refactor. I should probably start to work in a master/dev branch scenario. So no need since my next commit fixes it. But thanks.

PR #53, opened 2019-11-20
Update README.md

tcyrus:


makepaddev:
Thanks! but readme already diverging, we haven't 'exactly' gone alpha yet, but working on making the exact state more clear.

PR #87, opened 2020-06-20
Check in Cargo.lock for reproducible builds

teohhanhui:
Cargo.lock should be checked in for binaries - https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries


teohhanhui:
Problem discovered while I'm making an AUR package: https://aur.archlinux.org/packages/makepad-git

PR #94, opened 2020-08-31
microserde: parse tuple types in structures

dsp:
Tuple are valid types for a structures and enums in Rust. The current
implmentation only parses arrays. We extend the use of `eat_type` to
look for parenthesis '(' and parse it as a tuple. The implementatio has
been tested by extended the example case to include a tuple.

Feel free to ignore the pull request and reimplement this. I don't have a deep insight into proc macros and your implementation to ensure error conditions are okay. I am happy to also work on it more given some feedback on the change.


dsp:
I added a bunch of fixes for structs with anonymous fields. I can separate that into a different pull request if you want to, but opted to keep it as one for now to reduce merge conflicts.

dsp:
The CI breakage is unrelated.

PR #103, opened 2020-11-15
Update website demo link

DGriffin91:
I'm not sure if https://makepad.dev is down or just moved to https://makepad.nl.


makepaddev:
No it was just down, i restarted the server. Apparently it has a file handle leak somewhere.

PR #107, opened 2020-12-03
Remove duplicate entry in the toml file

cybear:
Fixes this error for macos:

```
~/c/makepad (master)> cargo run -p makepad --release
error: failed to parse manifest at `/Users/bjorn/code/makepad/render/Cargo.toml`

Caused by:
  could not parse input as TOML

Caused by:
  redefinition of table `target.aarch64-apple-darwin.dependencies` for key `target.aarch64-apple-darwin.dependencies` at line 26 column 1
```

### DIFF
--- a/render/Cargo.toml
+++ b/render/Cargo.toml
@@ -23,9 +23,6 @@ makepad-draw-derive = { path = "./draw_derive", version = "0.1" }
 [target.aarch64-apple-darwin.dependencies]
 makepad-objc-sys = { path = "./bind/objc-sys", version = "0.2" }
 
-[target.aarch64-apple-darwin.dependencies]
-makepad-objc-sys = { path = "./bind/objc-sys", version = "0.2" }
-
 [target.x86_64-apple-darwin.dependencies]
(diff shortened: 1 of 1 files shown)


makepaddev:
Thanks fixed it!

cybear:
Was there anything wrong with my PR that made you close it instead of merging?

makepaddev:
Oh no nothing wrong i just have a tendency if its a trivial change like deleting one liner to simply edit it in github UI instead of merging PR's. Maybe silly, just a habit.

cybear:
From a community perspective I think people would be happy to see their own names among the committers. But sure, this was a very small fix :)

makepaddev:
Yea no i get that, and for really nontrivial changes i definitely consider it, however from a copyright sanity and CLA point of view i wanna keep the tree clean of trivial things like renames/lineremoves or removing spaces and things like that.

makepaddev:
We're still contemplating what to do about CLA's. Most large more corporate projects have these contracts to make sure the (C) is unencumbered but that is signicant effort for the committer as well.

PR #117, opened 2021-04-28
Add middle click handling for tab

komonad:
I guess this will be a useful feature for tab management


makepaddev:
Hi sorry for completely ignoring this pull request. We aren't ready for pull requests at this point because our code is still massively in flux. Thanks for your effort tho, but please check back later as we stabilise more.

PR #124, opened 2022-01-22
Bug fixes to better enable third-party devs

survirtual:
Issue:

1) Running "cargo run --bin makepad_studio" had some hard-coded expectations about the directory it would exist in. 

2) File editing always looks for a live-code macro even in files that don't have that at all, causing a panic.

3) No way to retrieve solid components types to then edit parameters due to absence of downcasting on the trait object.

Fixes:

1) Retrieve the os working dir for the app and strip the prefix for files using that instead of hardcoded.

2) Added a check to see if the file has a live-code macro before running the live edit logic

3) Use the macro to generate a downcast for FrameComponent trait object, and modified the get_component macro to appropriately reflect changes.

Result:

These fixes let third party devs run the studio application to play with the app as expected and allows the modification of component props in code outside of the component, by being able to operate on the component directly.

### DIFF
--- a/component/src/frame_registry.rs
+++ b/component/src/frame_registry.rs
@@ -25,6 +25,8 @@ pub trait FrameComponent: LiveApply {
     fn type_id(&self) -> TypeId;
 }
 
+generate_ref_cast_api!(FrameComponent);
+
 #[derive(Clone)]
 pub struct FrameActionItem {
     pub id: LiveId,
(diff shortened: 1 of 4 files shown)


makepaddev:
Hi sorry for completely ignoring this pull request. We aren't ready for pull requests at this point because our code is still massively in flux. Thanks for your effort tho, but please check back later as we stabilise more. The reason i didnt have a get_component_mut is because you never get components as non mut. However i agree that this is nonstandard. I'll take it under advisement

PR #138, opened 2022-05-03
Shrink unsafe blocks for the ClassDecl/Method/Class impl

peamaeq:
In this impl you use the unsafe keyword for some safe expressions.
We need to mark unsafe operations more precisely using unsafe keyword. Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety  within an unsafe block. This is the balance between Safe and Unsafe Rust. The separation is designed to make using Safe Rust as ergonomic as possible, but requires extra effort and care when writing Unsafe Rust. 
Hope this PR can help you.
Best regards.
**References**
https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html 
https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html


makepaddev:
Hi sorry for completely ignoring this pull request. We aren't ready for pull requests at this point because our code is still massively in flux. Thanks for your effort tho. I agree unsafe should be used extremely sparingly and i have been reducing the use of unsafe already as much as i can in the platform layers. However i do sometimes use 'too big' of an unsafe block there which is probably better to reduce. Please check back when we do a more official release to see if i addressed enough of it but merging pullrequest for now is going to be hell.

Issue #156, opened 2023-03-24
Mobile keyboard support for Android


makepaddev:
Text input widget needs mobile keyboard support in the android backend
- pops up when focus 
- closes when focus lost 
- support clipboard
- supports swype/IME behavior

If you are interested in working on this problem contact us on discord so we can discuss architecture (see the main readme)

PR #161, opened 2023-03-24
Mobile keyboard support for Android 

jmbejar:
This PR tracks work related with #156

TODO list 
- [x] Display/hide keyboard when text input has/lost focus
- [x] Dispatch TextInput event when keys are pressed (basic support)
- [x] Support Shift modifier
- [x] Support all key codes (maybe creating a KeyCode constants for Android case instead of reusing the platform ones?)
- [x] Copy-Paste (clipboard) basic support
- [x] Cut (clipboard) basic support
- [x] Handle special keys: backspace, enter (dispatch appropriate events)
- [x] Avoid keyword to overlap focused text input
- [x] Capture KeyEvents modifiers (and check if this make keyword selection, using Shift, work)

Things to be considered in further PR
- Clipboard API access
- Mic icon (dictation) support


jmbejar:
Explanation about https://github.com/makepad/makepad/pull/161/commits/f4b07f3eaf743b66056344d5487a725ee1e70dc2 

When I was about to implement the contextual menu to show the option to copy selected text, I found that the text selection was discarded when the touch started (onTouchDown), in the context of the code of text_input.rs. I needed to make some changes to "reset the selection" in onTouchMove instead. 

As far I tested, the final result seems to be closer to existing behavior in text input controls from devices, where a simple tap deselects and set the cursor in the tapped position, but a long press does not discard selection.

tools/cargo_makepad/src/android/java/dev/makepad/android/MakepadActivity.java
```diff
@@ -227,8 +243,93 @@ public void swapBuffers() {
...
+            cutItem.setVisible(true);
+        }
+
+        ClipboardManager clipboard = (ClipboardManager) getSystemService(Context.CLIPBOARD_SERVICE);
+        MenuItem pasteItem = menu.findItem(R.id.menu_paste);
+        if (!(clipboard.hasPrimaryClip())) {
+            pasteItem.setVisible(false);
+        } else if (
+            !(clipboard.getPrimaryClipDescription().hasMimeType(ClipDescription.MIMETYPE_TEXT_PLAIN)) &&
+            !(clipboard.getPrimaryClipDescription().hasMimeType(ClipDescription.MIMETYPE_TEXT_HTML))
```
jmbejar: TODO: Change this conditional to be `a || b` instead of `!a && !b`

PR #175, opened 2023-05-18
Apply Clippy lints

Speykious:
I asked if I could make a PR on Clippy, I was told it was ok if it makes sense.

Let me know what you think of this. Besides the first commit (1df327af251b187450997e1d11ac3d70318ffba4), I made it so that each category of clippy lint is fixed on its own commit so that it can easily be reverted if you disagree with one.

If there's a disagreement with a lint, we can just explicitly allow it on the codebase (unless you don't want to clog the lib.rs files with a few one-line clippy macros...) - better not spend any time arguing with the linter if it's just meant to help code quality :)


Speykious:
There's this very useful lint about the complexity of the block of code for a condition inside an `if`:
> in an `if` condition, avoid complex blocks or closures with blocks; instead, move the block or closure higher and bind it with a `let`
> for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#blocks_in_if_conditions

which is highlighted here:
https://github.com/makepad/makepad/blob/2d71c8fbbf3588cb37a917f8f455f7688864a0ea/draw/vector/src/trapezoidator/mod.rs#L28-L52
but I have no idea what the condition should be called because I can't figure out what it checks. What would be a good name for it?

Speykious:
Now there are important lints, which in Clippy are marked as errors. For example:

> the trait `std::marker::Copy` cannot be implemented for this type

image

(...ok actually this one is straight up a Rust compiler error, not even a lint.)
In this case it's because `Rc<T>` doesn't implement Copy, so the enum shouldn't either. Would that make it too inconvenient?

***

> this comparison involving the minimum or maximum element for this type contains a case that is always true or always false
because `std::u32::MAX` is the maximum value for this type, this comparison is always true
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons

image

PR #177, opened 2023-05-19
Networking Layer - ChatGPT app

yulnr:
Doing:
- [x] Figure out best approach (how and where) for serializing request body
- [x] Complete conversions between HttpRequest and Responses Java <> Rust
- [x] Get it working grabbing a prompt from text input
   - [x]  add proper headers, etc.
- [x] Add live_id to request and response
- [x] Add error handling and HttpRequestError event 
- [x] Fix LiveId usage (maybe add a LiveId::from_u64)

We decided to tackle these in separate PRs:
- Websockets
For future iteration:
- Async
- Provide a more flexible API for response body serialization, allowing for using other than micro_serde and other than JSON

Findings
- micro_serde has an issue with trailing commas on json (fixed in this MR)
- micro_serde doesn't support a way to serialize into a generic struct, (feature from serde useful for handling response errors)
- micro_serde doens't support overriding field name mapping. (e.g. if a json object has a `type` field we need to write in Rust as `_type`, we currently don't have a way to tell micro_serde to map `type` -> `_type`)

^ these could not be a problem at all if we make it easy for users to use other serialization libs, however I'm assuming we want to provide the complete solution ourselves.


widgets/src/text_input.rs
```diff
@@ -736,4 +736,12 @@ impl TextInputRef {
...
         }
     }
+    
+    pub fn get_text(&self) -> String {
+        if let Some(inner) = self.borrow(){
+            inner.text.clone()
+        } else {
+            "".to_string()
+        }
+    }
```
yulnr, line 746: I'm unsure if this is a hack and we should use data binding instead.

PR #186, opened 2023-06-17
A quick fix to make Dock work in master

j16r:
Move the const definition to the outer scope


makepaddev:
Thx for the fix, i already changed it tho when you mentioned on discord.

PR #190, opened 2023-07-06
Extract image widget

jmbejar:
Extract a new `Image` widget based in the existing one that was implemented into `Frame` code.
It adds the some common transformations: rotation, opacity & scaling.

The legacy one is renamed to `ImageFrame` until we decide what it is the best way to refactor `Frame`.

Examples:

```
            // New Image widget
            img1 = <Image> {
                source: dep("crate://self/resources/tinrs.png")
                walk: {width: 100, height: 30},
                draw_bg: {
                    rotation: 1.0,
                    scale: 0.3,
                    opacity: 0.3,
                }
            }

            // Legacy way
            frm1 = <ImageFrame> {
                image: dep("crate://self/resources/tinrs.png")
                image_scale: 0.3,
                walk: {width: 100, height: 30},
            }
        }
```

### DIFF
--- a/widgets/src/frame.rs
+++ b/widgets/src/frame.rs
@@ -492,18 +492,6 @@ pub struct FrameRef(WidgetRef);
 pub struct FrameSet(WidgetSet);
 
 impl FrameRef {
-    /*
-    pub fn handle_event(&self, cx: &mut Cx, event: &Event) -> WidgetActions {
-        self.0.handle_widget_event(cx, event)
-    }
-    
-    pub fn draw(&self, cx: &mut Cx2d,) -> WidgetDraw {
-        if let Some(mut inner) = self.inner_mut() {
-            return inner.draw(cx)
-        }
-        WidgetDraw::done()
-    }*/
-    
     pub fn set_visible(&self, visible: bool) {
         if let Some(mut inner) = self.borrow_mut() {
             inner.visible = visible
--- a/widgets/src/frame.rs
+++ b/widgets/src/frame.rs
@@ -4,7 +4,7 @@ use {
         makepad_derive_widget::*,
         makepad_draw::*,
         widget::*,
-        widget_with_image::*,
+        image_loading_widget::*,
         scroll_bars::ScrollBars,
     },
 };
@@ -403,7 +403,7 @@ struct FrameTextureCache {
     color_texture: Texture,
 }
 
-impl WidgetWithImage for Frame {
+impl ImageLoadingWidget for Frame {
     fn image_filename(&self) -> &LiveDependency {
         &self.image
     }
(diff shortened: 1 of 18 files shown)


widgets/src/widget_with_image.rs
```diff
@@ -0,0 +1,74 @@
+use {
+    crate::{
+        makepad_image_formats::*,
+        makepad_draw::*,
+    }
+};
+
+pub trait WidgetWithImage {
+    fn image_filename(&self) -> &LiveDependency;
```
jutsuu, line 9: NIT: I think we could make it a bit clearer with making the name of the trait describe specifically what behaviour it adds to the implementer. 
Something like `ImageRenderingWidget` or `ImageLoadingWidget`

widgets/src/widget_with_image.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn texture(&mut self) -> &mut Option<Texture>;
```
yulnr: NIT: I think we could make it a bit clearer with making the name of the trait describe specifically what behaviour it adds to the implementer.
Something like `ImageRenderingWidget` or `ImageLoadingWidget`, but I know those aren't amazing either.

widgets/src/frame.rs
```diff
@@ -741,10 +711,6 @@ impl Frame {
         self.area
     }
     
-    pub fn draw(&mut self, cx: &mut Cx2d,) -> WidgetDraw {
-        self.draw_walk(cx, self.get_walk())
-    }
-
```
yulnr, line 749: I believe that we're removing this because the code using it in FrameRef is commented? Should we either delete that code also or keep both?

widgets/src/image.rs
```diff
@@ -0,0 +1,174 @@
...
+            // In any case, it can be override by the app developer.
+            clip_x: false,
+            clip_y: false
+        }
+        draw_bg: {
+            texture image: texture2d
+
+            instance rotation: 0.0
+            instance opacity: 1.0
+            instance scale: 1.0
```
yulnr, line 23: Haven't fully gone through the rotation changes, but should we expect to use these instances through overriding `draw_bg ` and have it working with the current code?

I tried this:

```rs
  <Image> {
      source: dep("crate://self/resources/tinrs.png"),
      walk: {width: (1000 * 0.175), height: (175 * 0.175), margin: 0}
      layout: {padding: 0}
      draw_bg: {
          scale: 1.0
          // rotation: 2.0
      }
  }
```

`scale` works as expected but if I use any `rotation` the image stops displaying, (tried this with the tinrs.png in the ironfish demo). Investigating in case I've broken something, or to figure out why it is happening

widgets/src/frame.rs
jmbejar: Good call.. I'm removing the related commented code

widgets/src/image.rs
jmbejar: This issue should be fixed now.

yulnr (approved):

PR #195, opened 2023-08-03
draw: integrate `rustybuzz` for shaping text.

eddyb:
Opening as draft ~~because of the example commit, but also~~ because this is rebased on top of:
* #188

For a more standalone/self-contained code sample of using the same APIs, see:
* https://github.com/RazrFalcon/rustybuzz/pull/70

---

The example isn't the best for this (e.g. multi-line text would be ideal), but:
|**Before**||
|-:|-|
|**After**||

<sub>(Chinese being correct without shaping isn't surprising, but Hindi can be impacted, this single word just doesn't happen to do anything interesting, but finding a simple example that's both realistic and insightful, is hard)</sub>


eddyb:
I've split the [test commit into its own branch](https://github.com/eddyb/makepad/compare/rustybuzz...eddyb:makepad:rustybuzz-example) so this can be landed on its own.

PR #200, opened 2023-08-15
Adjust LiveView widget flick scrolling to behave more natural.

jmbejar:
After comparing with native apps in Android, we changed a few things.

We now differentiate between a "short drag", being the one that defines the swipe movement for the consequent scroll of the list, from a "long drag".

The latter causes the scroll animation to be stopped and, as a consequence, the list view is scrolled directly by the drag motion as it is done by the user.

### DIFF
--- a/examples/news_feed/src/app.rs
+++ b/examples/news_feed/src/app.rs
@@ -243,7 +243,7 @@ live_design!{
         layout: {flow: Down, padding: 0.0, spacing: 0.0}
         
         hero = <Image> {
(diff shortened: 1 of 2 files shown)


univerz:
btw @jmbejar are you on druid/xilem zulip? there is a new discussion about [kinetic scrolling](https://xi.zulipchat.com/#narrow/stream/351333-glazier/topic/Kinetic.20and.20line.2Fpage.20scrolling/near/387346577) that you might be interested in.

PR #239, opened 2023-09-12
Prompt user for iOS build settings

cornedriesprong:
if there are multiple options, prompt user for:
- signing identity
- provisioning profile
- target device


makepaddev (commented):
Btw could you instead of using an interactive prompt just dump the list and tell people to pick an identity number on the commandline ? like --identity=x that way we can integrate it into a buildtool. Asking on stdin is blocking automated tools.

cornedriesprong:
@makepaddev updated

PR #282, opened 2023-09-22
linux_direct can now use input devices that are plugged in after the application has started

SpieringsAE:
Moved the event gathering/grouping into the event threads so poll_raw_input can never block.
Improved the for i in 1..12 event# file reading to a specific glob like function.

Currently the new inotify dependency in libs is on all Linux systems, It seems impossible to limit this to cfg(linux_direct).
In the Cargo.toml files only the standard cfgs are available, like target_os and such (rustc --print=cfg).
The build.rs script runs after all depencies are compiled so this also isn't an option.
Can't use features because the dependency on makepad-platform is several layers deep, this would mean every package that depends on makepad platform would also have to add the features, and then those that depend on those aswell. Making for a big ass mess.

This leaves only 2 options to make this work:
1. Custom targets for linux_direct, but this sounds like hell.
2. Moving libs/inotify-rs into platform/src/os/linux/direct and dont put it in Cargo.toml at all, but this is ugly and ruins the structure.


SpieringsAE:
Failed a bit with this pull request, here is some extra info: \
linux_direct target now has to be set with the RUSTFLAGS env variable unfortunately, untill rust fixes this build issue:
```
RUSTFLAGS="--cfg linx_direct" cargo build etc.
```
The pre_push_check.sh script commandos for checking linux_direct have been adapted to this.

Now supports a Type B MultiTouch device with the makepad touch api:
https://www.kernel.org/doc/Documentation/input/multi-touch-protocol.txt

Input event gathering is now done off the main thread so that poll_raw_input() will never be able to block.

SpieringsAE:
a large push has happened in the mean time. Made my own version of the libevdev library to handle raw input devices. This enables far greater precision when handling input events by check device properties/available types and codes etc.

makepaddev:
Why did you do this part:
MAKEPAD=lines RUSTFLAGS="--cfg linux_direct" cargo +nightly check -q -p makepad-example-ironfish --release --target=x86_64-unknown-linux-gnu --message-format=json
Does MAKEPAD=linux,linux_direct not work?

makepaddev:
Sorry looks like i split on +. So this
Does MAKEPAD=lines+linux_direct not work?

SpieringsAE:
Because the dependency on inotify depends on --cfg linux_direct.
build.rs gets resolved after all the depencies are resolved, so it would not depend on inotify, but would try to compile code that is gated behind the the cfg(linux_direct) causing the build to fail.

The only way to solve this without adding RUSTFLAGS="--cfg linux_direct", is by making all the inotify dependency lean on all linux targets for example, even though the normal linux target doesnt use this dependency. This is an issue that has been raised see: 
https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618
But is currently not resolved.

So it can be returned to the old format now, but it would cause every linux target to depend on something that it might not actually depend on.

One other way is to move the inotify lib into the platform crate, so it can be optionally compiled with cfg(linux_direct) but I feel like this would be very messy, and also disallow other crates from maybe adapting this lib.

So yeah there is 4 options:
1. Leave it like this (which I don't really like either)
2. Have all Linux targets depend on it like I did it in the beginning (probably gets filtered out in the linking step but still generates extra compile time)
3. Move inotify into the platform crate
4. Don't listen for devices that connect after the app is started and remove inotify completely. I personally really like being able to plug and unplug my devices while the app is running, but in a final/production linux_direct app I guess this scenario is unlikely.

All of these are of course only valid until the previously mentioned issue is resolved by rust itself, making room for a cleaner solution.

makepaddev:
We're going to merge inotify into makepad-platform first, since we need filewatchers all over the place

makepaddev:
Not as a crate but just as a module inside makepad platform

SpieringsAE:
okay that sounds great, I kind of expected it to become useful in the project. Once that has happened I will adapt these changes to that approach so the weird RUSTFLAGS="--cfg linux_direct" can be removed hopefully.

SpieringsAE:
It seems like this request isnt going to come through in its current for, I will close it for now and make a request into a dev branch once I have improved some parts about this

PR #285, opened 2023-09-28
Implement editor navigation:

fjakobs:
- goto line start/end (Cmd-left, Cmd-right)
- select all (Cmd-A)

many more are missing


fjakobs:
cc @nightwing

PR #320, opened 2023-11-12
fixed the problem of cursor setting after window startup and fix string unicode escaping issue so that it can be recognized correctly when using custom fonts.

get200:


get200:
## 1.fixed the problem of cursor setting after window startup
### before
mouse_move_in_before
### after
mouse_move_in_after

## 2.fix string unicode escaping issue
### custom font




### before


### after

PR #346, opened 2023-12-15
Make texture pool format-aware on re-usage

yulnr:
Fixes font textures breaking on taobao app when using video. 
Currently all Makepad textures are re-used indiscriminately, which breaks in OpenGL when trying to perform `TEXTURE_2D` operations on a previous texture bound with `TEXTURE_EXTERNAL_OES` format.
This PR makes the texture pool check for format compatibility when re-using textures from the free pool.

Changes:
- Extends `IdPool` with an `alloc_with_reuse_filter` method that takes in a filtering callback.
-`Texture::new` now always creates a default black texture, and `texture.set_format(cx, format)` no longer exists. If format (even size alone) needs to be changed, a new texture must be created with `Texture::new_with_format(cx, format)`
- Deletes OS texture handles and re-creates new ones when texture is re-used (generation changed). Added this on OpenGL and Metal, DirectX is already re-creating and I'll update WebGL once we can fully test it.

### DIFF
--- a/audio_graph/audio_widgets/src/display_audio.rs
+++ b/audio_graph/audio_widgets/src/display_audio.rs
@@ -94,7 +94,7 @@ const WAVE_SIZE_Y: usize = 16;
 impl LiveHook for DisplayAudio {
 
     fn after_new_from_doc(&mut self, cx: &mut Cx) {
-        self.wave_texture.set_format(cx, TextureFormat::VecBGRAu8_32 {
+        self.wave_texture = Texture::new_with_format(cx, TextureFormat::VecBGRAu8_32 {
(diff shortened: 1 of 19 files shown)


univerz:
probably related - makepad-example-news-feed also shows listing of items without text on android - without using video, works fine on linux. the same with your branch.

yulnr:
> probably related - makepad-example-news-feed also shows listing of items without text on android - without using video, works fine on linux. the same with your branch.

thanks for mentioning! I'm checking for other side effects, so I'll investigate. Might open a new issue and if turns out to be unrelated 🙌🏽

PR #356, opened 2024-01-03
Fix texture pool not updating inner item on re-use

yulnr:
Fixes a bug where setting images within a list of multiple images with `load_png_from_bytes()`, causes the most recently-loaded image to be re-used all other images in the list. 

Screenshot 2024-01-03 at 16 03 31

The problem is that textures won't update their contents/format when being re-used with `new_with_format` because the texture pool (`IdPool`) is not updating the inner item when providing textures from the free pool.

Updating the inner item (`self.pool[id].item = item`) fixes the issue. 

image

Also, this removes the need for the `generation_changed` flag. Since we're now properly updating the contents of the texture, `alloc` will be set as `None` and therefore always require the graphics backend to regenerate textures.

Originally reported here: https://github.com/project-robius/robrix/commit/5a79e97cfc78a12d221918fb77a8e50a768fa6ae

### DIFF
--- a/platform/src/id_pool.rs
+++ b/platform/src/id_pool.rs
@@ -85,8 +85,9 @@ impl<T> IdPool<T> where T: Default {
             });
     
         if let Some((index, id)) = maybe_free_id {
-            self.pool[id].generation += 1;
             self.free.0.borrow_mut().remove(index);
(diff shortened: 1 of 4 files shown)


kevinaboos:
Thanks Julián!

PR #364, opened 2024-01-10
Fix typo in cfg block

kevinaboos:
tiny fix to allow `rik` branch to build again


kevinaboos:
Ah, #363 beat me to it

PR #366, opened 2024-01-16
Explicitly return a `Result` when an image cannot be loaded

kevinaboos:
The `ImageRef::load_*` functions currently just print errors to stdio instead of returning them, making it impossible to programmatically determine whether an image has failed to load.

This PR changes all image-related functions to explicitly return errors that could occur internally, via a new error enum:
```rust
pub enum ImageError {
    /// The image data buffer was empty.
    EmptyData,
    /// The image's pixel data was not aligned to 3-byte or 4-byte pixels.
    InvalidPixelAlignment,
    /// The image data could not be decoded as a JPEG.
    JpgDecode(JpgDecodeErrors),
    /// The image file at the given resource path could not be found.
    PathNotFound(String),
    /// The image data could not be decoded as a PNG.
    PngDecode(PngDecodeErrors),
    /// The image data was in an unsupported format.
    /// Currently, only JPEG and PNG are supported.
    UnsupportedFormat,
}
```

To preserve backwards compatibility, I silently drop that `Result` in a few places where there is currently no way to return it, e.g., in a few `LiveApply` trait impls.

### DIFF
--- a/widgets/src/image.rs
+++ b/widgets/src/image.rs
@@ -4,6 +4,7 @@ use crate::{
     makepad_draw::*,
     widget::*
 };
+use std::fmt;
 
 live_design!{
     ImageBase = {{Image}} {}
@@ -36,7 +37,7 @@ impl LiveHook for Image{
         self.lazy_create_image_cache(cx);
         let source = self.source.clone();
         if source.as_str().len()>0 {
-            self.load_image_dep_by_path(cx, source.as_str())
+            let _ = self.load_image_dep_by_path(cx, source.as_str());
         }
     }
 }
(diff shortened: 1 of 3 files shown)


kevinaboos:
So far I have only tested this on macOS with `cargo +nightly build --all`, and double checked that none of the many warnings were related to this. 

I'm not sure how to run the local test suite that covers all other platforms, but please let me know how to do that so I can make sure it doesn't break makepad on another platform.

widgets/src/image.rs
```diff
@@ -101,21 +102,30 @@ impl Image {
 }
 
 impl ImageRef {
-    pub fn load_image_dep_by_path(&self, cx: &mut Cx, image_path: &str) {
+    /// Loads the image at the given `image_path` into this `ImageRef`.
+    pub fn load_image_dep_by_path(&self, cx: &mut Cx, image_path: &str) -> Result<(), ImageError> {
         if let Some(mut inner) = self.borrow_mut() {
             inner.load_image_dep_by_path(cx, image_path)
+        } else {
+            Ok(()) // preserving existing behavior of silent failures.
```
univerz, line 110: i think @makepaddev  had some argument for not having an `Result` in the rendering part of the api, but from my point of view - it's better to fail sooner rather than debug silent errors. + it would remove huge amount of unwraps from the codebase.

yulnr:
> I'm not sure how to run the local test suite that covers all other platforms, but please let me know how to do that so I can make sure it doesn't break makepad on another platform.

@kevinaboos historically we ran the script: `tools/pre_push_check.sh`. IIRC the web build might not be compiling right now so you can ignore that one. 
Nowadays you can also run the studio (`cargo run -p makepad-studio`) and click Check `All` (or a specific platform) under `cargo-makepad`.  Or you can also just compile a specific project for the different platforms if you don't want to DDoS your own CPU as much 😅

image

I just saw that `cargo makepad check all` is documented in the readme but not currently working. I'll investigate

widgets/src/image.rs
kevinaboos: I do agree with that general philosophy, however that would be a _massive_ change in behavior across the whole repo, and likely isn't something you'd want to have to deal with here or would even be able to handle properly.

All of the `WidgetRef` types use this pattern of doing nothing if the inner reference cannot be obtained, so I don't think it's appropriate to change it here for just a single `WidgetRef` type.

kevinaboos:
> historically we ran the script: `tools/pre_push_check.sh`. IIRC the web build might not be compiling right now so you can ignore that one.

Thanks, just ran that and it worked fine AFAICT.

> Nowadays you can also run the studio (`cargo run -p makepad-studio`) and click Check `All` (or a specific platform) under `cargo-makepad`. Or you can also just compile a specific project for the different platforms if you don't want to DDoS your own CPU as much 😅

Ah cool, I'm not using makepad studio daily yet, so I didn't know about that. I get loads of errors when I run each of those, but the PR was merged so I assume Rik checked it locally and deemed it okay.

PR #367, opened 2024-01-16
Provide more details about an invalid pixel alignment error

kevinaboos:
Also removes a very verbose log statement inadvertently added in d8738bd


kevinaboos:
resolved conflicts again.

PR #369, opened 2024-01-20
Retrieve and use the proper mouse button number for other mouse button clicks (macOS only)

kevinaboos:
* Currently, the button numbers in use are based off of the `buttonNumber` property from macOS `NSEvent` API: <https://developer.apple.com/documentation/appkit/nsevent/1527828-buttonnumber>
  * 0: Main (left) mouse button
  * 1: Secondary (right) mouse button
  * 2: Middle mouse button (clicking the scroll wheel)
  * 3: Fourth mouse button, commonly the "back" button
  * 4: Fifth mouse button, commonly the "forward" button

-------------

As I said on Discord, I'm not sure if any Makepad widget impls or apps actually depend on the previous simplified behavior of always using `2` for any other mouse button press.

If so, I can augment this PR with a MouseButton enum in `platform/events` that can be used across Makepad such that we discourage usage of raw integers.

### DIFF
--- a/platform/src/os/apple/apple_util.rs
+++ b/platform/src/os/apple/apple_util.rs
@@ -406,6 +406,19 @@ pub fn bottom_left_to_top_left(rect: NSRect) -> f64 {
     height as f64 - (rect.origin.y + rect.size.height)
 }
 
+/// Retrieves the mouse button number for the given `event.
+///
+/// Button numbers are as follows:
+/// * 0: Main (left) mouse button
+/// * 1: Secondary (right) mouse button
+/// * 2: Middle mouse button (clicking the scroll wheel)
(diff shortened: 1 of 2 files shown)


kevinaboos:
Also, according to [the docs](https://developer.apple.com/documentation/appkit/nsevent/1527828-buttonnumber?language=objc), we can actually use this for all mouse events. So we could also simplify the code in `macos_delegates.rs` to reduce 6 functions into 2:
* `mouse_down()`, `right_mouse_down()`, and `other_mouse_down()` into a single function
* `mouse_up()`, `right_mouse_up()`, and `other_mouse_up()` into a single function

Let me know if you'd like me to make that change too.

PR #388, opened 2024-02-02
`cargo-makepad`: support installing the full Android NDK (macOS only)

kevinaboos:
Currently this is macOS only because that's the only machine I have to test it on right now.

Also includes the following related changes:

* Use the canonical directory structure for the NDK toolchain path.
  * This enables a user of `cargo-makepad` to point to a separate regular installation
    of the Android SDK, e.g., one installed automatically via Android Studio's SDK manager.
* Export three environment variables to make native code builds "just work"® with the `cc` crate:
  * "CC_{target-triple}"
  * "AR_{target-triple}"
  * "RANLIB_{target-triple}"

### DIFF
--- a/tools/cargo_makepad/src/android/sdk.rs
+++ b/tools/cargo_makepad/src/android/sdk.rs
@@ -199,8 +199,7 @@ pub fn expand_sdk(sdk_dir: &Path, host_os: HostOs, args: &[String], targets:&[An
             let copy_fn = if full_ndk { shell::cp_all } else { shell::cp };
             copy_fn(&mount_point.join(source_path), &sdk_dir.join(dest_path), *exec) ?;
         }
-        // TODO: restore before merging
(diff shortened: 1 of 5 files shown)


tools/cargo_makepad/src/android/sdk.rs
```diff
@@ -324,44 +330,66 @@ pub fn expand_sdk(sdk_dir: &Path, host_os: HostOs, _args: &[String], targets:&[A
...
+                        .filter_map(|r| r.ok().and_then(|entry| {
+                            let path = entry.path();
+                            path.is_file().then_some(path)
+                        }))
+                        .for_each(|bin_file|
+                            std::fs::set_permissions(&bin_file, PermissionsExt::from_mode(0o744))
+                                .expect("failed to set exec permissions on {bin_file:?}")
+                        );
+                }
+            } else {
```
kevinaboos, line 355: all of the changes from here down are just indentation

PR #394, opened 2024-02-09
Print warning about missing templates for list-like widgets (FlatList, PortalList, Dock)

kevinaboos:
I constantly seem to forget to add newly-created widget templates to lists, and I assume a lot of other devs also make this mistake. This results in me always taking a few minutes to remember where I went wrong and what I need to do to fix it.

This small PR improves that dev experience by printing out a warning about the missing template with a suggestion of how to fix it. Hopefully will result in less wasted time tracking down silly absent-minded mistakes like this.

Ideally we would be able to detect this at compile time, but that'd be quite a major redesign of the live_design system. Probably best left to an LSP + IDE plugin later.

### DIFF
--- a/widgets/src/dock.rs
+++ b/widgets/src/dock.rs
@@ -524,12 +524,12 @@ impl Dock {
             let entry = self.items.get_or_insert(cx, entry_id, | cx | {
                 (template, WidgetRef::new_from_ptr(cx, Some(*ptr)))
             });
-            return Some(entry.1.clone())
+            Some(entry.1.clone())
         }
         else {
-            log!("PortalList template not found {}", template);
+            warning!("Template not found: {template}. Did you add it to the <Dock> instance in `live_design!{{}}`?");
+            None
         }
-        None
     }
     
     pub fn items(&mut self) -> &ComponentMap<LiveId, (LiveId, WidgetRef)> {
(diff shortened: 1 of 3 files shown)


widgets/src/dock.rs
```diff
@@ -524,12 +524,12 @@ impl Dock {
             let entry = self.items.get_or_insert(cx, entry_id, | cx | {
                 (template, WidgetRef::new_from_ptr(cx, Some(*ptr)))
             });
-            return Some(entry.1.clone())
+            Some(entry.1.clone())
         }
         else {
-            log!("PortalList template not found {}", template);
+            warning!("Template not found: {template}. Did you add it to the <Dock> instance in `live_design!{{}}`?");
```
univerz, line 530: btw @makepaddev - it would be great to rename this to `warn!` to make it compatible with other rust log implementations. or use one of them and add callback to things like adb, so we can collect logs from external crates too?

a log callback on makepad side would definitely be useful to allow logs to be exported from the user's device.

kevinaboos: Yeah I'll probably eventually work with the makepad team to implement a `log`-compatible and `tracing`-compatible backend such that other crates in the dependency chain can also output to it.

PR #400, opened 2024-02-18
fix linux timers compiled with opt-level > 0

univerz:
`as_mut()` is the only important part, the rest is just cleanup. this is another one of my "fixes without deep understanding" => take it with a grain of salt, but it makes sense.

### DIFF
--- a/platform/src/os/linux/select_timer.rs
+++ b/platform/src/os/linux/select_timer.rs
@@ -1,15 +1,6 @@
-use {
-    std::{
-        collections::{VecDeque},
-        time::Instant,
-        mem,
-        os::raw::{c_int},
-        ptr,
-    },
-    self::super::{
-        libc_sys,
-    },
-};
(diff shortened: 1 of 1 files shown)


univerz:
without fix `--release` build stops firing timer event & `Event::Signal` doesn't work without UI input (mouse, ...)

```diff --git a/platform/src/os/linux/x11/linux_x11.rs b/platform/src/os/linux/x11/linux_x11.rs
index cf132ad0..c05612df 100644
--- a/platform/src/os/linux/x11/linux_x11.rs
+++ b/platform/src/os/linux/x11/linux_x11.rs
@@ -192,7 +192,7 @@ impl Cx {
                 self.call_event_handler(&Event::TextCut(e))
             }
             XlibEvent::Timer(e) => {
-                //println!("TIMER! {:?}", std::time::Instant::now());
+                println!("TIMER! {:?}", std::time::Instant::now());
                 if e.timer_id == 0{
                     if SignalToUI::check_and_clear_ui_signal(){
                         self.handle_media_signals();
```

PR #403, opened 2024-02-21
support for Android back button onBackPress

guofoo:
handle_back_pressed() will be called when the Android back button is pressed


univerz:
thank you :)

PR #409, opened 2024-02-26
glsl: switch to `#version 300 es` and use UBOs (uniform buffers) via "uniform blocks".

eddyb:
Pros:
- `#version 300 es` fully bypasses the Adreno driver bugs worked around recently
  - see also #360
  - [`OES_EGL_image_external`](https://registry.khronos.org/OpenGL/extensions/OES/OES_EGL_image_external.txt) is superseded by [`OES_EGL_image_external_essl3`](https://registry.khronos.org/OpenGL/extensions/OES/OES_EGL_image_external_essl3.txt) and only the former seems to have buggy implementations
  - I've kept the bypass for Android emulators but maybe that's _also_ unnecessary?
- modern GLSL is required for some experimentation with tools like Naga (which is where this all started)
- relying on UBOs mean the OpenGL (ES) backend looks a lot more like the D3D11 one
  - refactoring all backends to centralize buffer management may be possible but hard to test atm
- more importantly, the OpenGL (ES) backend was the only one *not* using buffers for uniforms yet
  - any `wgpu` backend will require explicit buffers, so aligning all backends first helps
- `layout(std140)` also aligns with WebGPU/WGSL requirements around UBOs (D3D11 appears to also need this)

Cons:
- `layout(std140)` is _deeply_ strange, and designed around hardware that seems to prefer all memory to be arranged into units of `f32x4` (128-bit-word-addressable memory, if you will. possibly some kind of "constant cache")
  - however, it seems that HLSL/D3D11 also exposes this constraint, and some code (for `const_table`) is reusable
  - I'm not even sure why Metal doesn't care, last I checked it mattered on older Metal platforms?
  - despite everything available on Vulkan, this would still be needed on WebGPU/WGSL (see last "Pros" entry)
- OpenGL (ES) backend minimum requirements jump from OpenGL ES 2.x to OpenGL ES 3.0 (roughly OpenGL 3.3)
  - on the web, this is comparable to going from WebGL to WebGL2
  - as prior art, [`wgpu` backends](https://github.com/gfx-rs/wgpu#supported-platforms) include a `GLES 3.0 / GL 3.3 / WebGL2` combined backend
    - so in a `wgpu`-only world, this level of GL(ES) _would_ be the worst-case minimum support
    - crucially, that can lack compute shaders, but otherwise be able to run a subset of WebGPU/WGSL

---

This PR is split into independently reviewable commits, and each were tested against several Project Robius apps (~8 atm, could be more if I spent more time fixing unrelated pre-existing build failures).

Also, (less exhaustive) Android testing is what led to finding out `#version 300 es`/`OES_EGL_image_external_essl3` do not run into the same Adreno bug that `#version 100`/`OES_EGL_image_external` did.


eddyb:
(Temporarily in draft mode to allow investigating the WebGL1 situation - it should be possible to preserve compatibility with it, and only use UBOs in the "native GL" backend)

PR #420, opened 2024-03-06
Fix issue with TextInput cursor position

jmbejar:


univerz:
nice, this solves one of my issues :). do you plan to work on ligatures as well (try placing your cursor in `ffff fff ffff`)? 

btw, there is (probably) a similar problem when you type `aaarbbb` and double-click to select it (notice 2 selection ranges)
Screenshot_20240306-204515

this part of the code could use a bigger refactor.

(+ it would be great to move selection & copy-paste to `Label`)

PR #422, opened 2024-03-08
Fix #392. Broken Linux cursor

### LINKED ISSUE #392: platform: linux resize cursor remains when moving into window
Moving the mouse into a makepad window results in the resize cursor being left in place:

Screenshot from 2024-02-06 23-17-15

FWIW - Motion notify events call `set_mouse_cursor` which updates `self.current_cursor` and sets the x11 cursor.

https://github.com/makepad/makepad/blob/69bef6bab686284e1e3ab83ee803f29c5c9f40e5/platform/src/os/linux/x11/xlib_app.rs#L245
https://github.com/makepad/makepad/blob/69bef6bab686284e1e3ab83ee803f29c5c9f40e5/platform/src/os/linux/x11/xlib_app.rs#L279-L282
https://github.com/makepad/makepad/blob/69bef6bab686284e1e3ab83ee803f29c5c9f40e5/platform/src/os/linux/x11/xlib_app.rs#L640-L642

Perhaps if `set_mouse_cursor` could distinguish these resize cursor changes from other cursor changes it could skip updating `current_cursor` for resizing and only set the x11 cursor. Then the final else for the motion notify processing could set the x11 cursor back to `current_cursor`. 

https://github.com/makepad/makepad/blob/69bef6bab686284e1e3ab83ee803f29c5c9f40e5/platform/src/os/linux/x11/xlib_app.rs#L311-L320

mneilly:
This is a fix for #392 where the resize cursor persists when entering a Linux window. This change adds an 'internal' cursor which the Linux code uses to display the resize cursor and it only alters the current cursor when _set_mouse_cursor_ is called. When the resize region is entered the _internal_cursor_ is used and when the resize region is exited_ the _current_cursor_ is restored.

### DIFF
--- a/platform/src/os/linux/x11/xlib_app.rs
+++ b/platform/src/os/linux/x11/xlib_app.rs
@@ -54,6 +54,7 @@ pub struct XlibApp {
     //pub free_timers: Vec<usize>,
     pub event_flow: EventFlow,
     pub current_cursor: MouseCursor,
+    pub internal_cursor: MouseCursor,
     pub atoms: XlibAtoms,
     pub dnd: Dnd,
 }
@@ -84,6 +85,7 @@ impl XlibApp {
                 event_flow: EventFlow::Poll,
                 //free_timers: Vec::new(),
                 current_cursor: MouseCursor::Default,
+                internal_cursor: MouseCursor::Default,
                 dnd: Dnd::new(display),
             }
         }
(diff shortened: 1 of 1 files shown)


HookedBehemoth:
This commit makes strace very spammy as restore_mouse_cursor is called every frame where the cursor is not in a resize area.
Checking internal_cursor against current_cursor is one thing way to fix this, I suppose. Caching the cursors would also be nice.

PR #423, opened 2024-03-08
Install android SDK components in a canonical path that mimics Android Studio

kevinaboos:
Currently the proper paths are only set up canonically for builds on macOS. This will be extended to other host platforms in the near future.

* Export more environment variables for Rust builds in order to support building third-party crates that need to know details about the Android SDK.

* Expose the internal reference to the Android JVM and native main Activity such that they can be passed to third-party crates when needed.

-------------

With these changes, the `robius-authentication` crate can now be built by `cargo-makepad` and used by a makepad app!

### DIFF
--- a/tools/cargo_makepad/src/android/compile.rs
+++ b/tools/cargo_makepad/src/android/compile.rs
@@ -56,6 +56,7 @@ fn manifest_xml(label:&str, class_name:&str, url:&str)->String{
         <uses-permission android:name="android.permission.BLUETOOTH_CONNECT"/>
         <uses-permission android:name="android.permission.CAMERA"/>
         <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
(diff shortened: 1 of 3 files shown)


tools/cargo_makepad/src/android/compile.rs
```diff
@@ -56,6 +56,7 @@ fn manifest_xml(label:&str, class_name:&str, url:&str)->String{
         <uses-permission android:name="android.permission.BLUETOOTH_CONNECT"/>
         <uses-permission android:name="android.permission.CAMERA"/>
         <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
+        <uses-permission android:name="android.permission.USE_BIOMETRIC" />
```
kevinaboos, line 59: this is only here since we don't yet have a standardized way for a third-party app to add permissions

PR #431, opened 2024-03-14
Refactor `jni_sys` module out into a separate crate `libs/jni-sys/`

kevinaboos:
This enables us to use a `#[patch]` directive in an app's Cargo.toml file to swap out the upstream `jni-rs/jni-sys` crate with makepad's internal minified copy of `jni-sys`, or vice versa.


kevinaboos:
Note that I have tested this with several Android makepad apps but am still not 100% sure how to properly run a broader CI test, as some of the cargo check tests fail.

PR #441, opened 2024-04-10
html_experiment: hyphenated attributes not working (`data-mx-color`)

kevinaboos:
In `html_experiment`, the first two lines of `body` work, as they test the default `color` attributes.

However, the third line starting with `"testing span with colors"` does not work, as those attributes have hyphens.... at least I _think_ that may be the cause.

Similarly, the fourth line `"testing spoiler"` doesn't yet work, but that's okay because i haven't implemented the click handler for it. However, its `data-mx-spoiler` attribute also doesn't get parsed properly.

### DIFF
--- a/experiments/html_experiment/src/app.rs
+++ b/experiments/html_experiment/src/app.rs
@@ -32,6 +32,22 @@ live_design!{
         }
     }
 
+    // This is an HTML subwidget used to handle `<font>` and `<span>` tags,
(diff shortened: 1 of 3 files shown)


kevinaboos:
Closing due to hyphenated live IDs being supported in af1676d

kevinaboos:
Closing due to hyphenated live IDs being supported in af1676d

PR #444, opened 2024-04-11
Additions in some widgets for Moxin specific needs

jmbejar:
This PR adds:

* A flag to indicate if `DropDown` is open in shader code
* A public function to check if a `TextInput::Return` event was emitted
* New actions emitted from the `View` widget related with focus events

### DIFF
--- a/widgets/src/text_input.rs
+++ b/widgets/src/text_input.rs
@@ -650,6 +650,13 @@ impl TextInputRef {
         }
         None
     }
+
+    pub fn returned(&self, actions: &Actions) -> Option<String> {
+        if let TextInputAction::Return(val) = actions.find_widget_action_cast(self.widget_uid()) {
+            return Some(val);
+        }
+        None
+    }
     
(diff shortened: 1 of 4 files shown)


widgets/src/text_input.rs
```diff
@@ -650,6 +650,13 @@ impl TextInputRef {
...
     }
+
+    pub fn returned(&self, actions: &Actions) -> Option<String> {
+        if let TextInputAction::Return(val) = actions.find_widget_action_cast(self.widget_uid()) {
+            return Some(val);
+        }
+        None
+    }
     
     pub fn return_key(&self, actions: &Actions) -> Option<String> {
```
univerz, line 661: what is the plan with this function (it has the same implementation as the new one)?

jmbejar: Ahh good call.. two people adding the same function at almost the same time. I will discuss with @makepaddev to keep only one version

PR #459, opened 2024-04-28
Refactor Android context states (JavaVM, current Activity) into a separate crate

kevinaboos:
* This enables an external crate to more efficiently access those two Android states that are owned and managed directly by Makepad.

* The new `makepad-android-state` crate ensures that only the internal framework code within Makepad has the ability to set the Activity instance.

* The JavaVM instance is set automatically from the JNI layer, as before.


libs/android_state/README.md
```diff
@@ -0,0 +1,30 @@
...
+    to obtain the private "set_activity" function, which ensures that
+    only the internal Makepad framework can set the activity instance.
+
+## Usage
+> Note: you probably want to use the [`robius-android-env`] crate instead of using this crate directly, or an even higher-level crate that depends on [`robius-android-env`].
+
+External users of this crate should only care about two functions:
+1. [`get_java_vm()`]: returns a pointer to the JavaVM instance,
+   through which you can obtain the JNI environment.
+2. [`get_activity()`]: returns a pointer to the current Makepad Activity instance.
```
MarijnS95, line 25: These link targets aren't defined in the context of `README.md` :sweat_smile:: https://crates.io/crates/makepad-android-state/0.1.0

kevinaboos: Haha yep, we rushed to publish this before RustNL but I couldn't get the other crates published so it didn't matter. Will fix in a future round of PRs.

PR #463, opened 2024-05-08
x11: handle VisibilityNotify event and redraw

harryhaaren:
DWM window manager uses "Geom" changes on view/update, which does not cause a full redraw of the application before this diff. That results in a blank/old-garbage-content texture, instead of the Makepad UI. This fix simply asks the UI to do a full redraw.

### DIFF
--- a/platform/src/os/linux/x11/x11_sys.rs
+++ b/platform/src/os/linux/x11/x11_sys.rs
@@ -82,6 +82,13 @@ pub const EnterWindowMask: u32 = 16;
 pub const LeaveWindowMask: u32 = 32;
 pub const XBufferOverflow: i32 = -1;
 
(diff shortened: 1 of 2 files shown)


harryhaaren:
Found a better fix, which is not window manager specific (like this fix was, it was a hacky/workaround for "redraw on certain scenarios"). Future fix will use appropriate X11 event `VisibilityNotify` which was not in the generated bindings, and some states for that event type `Unobscured`, `PartiallyObscured` and `FullyObscured` to identify when to redraw.

harryhaaren:
Tested on `dwm` window manager (change virtual desktop "tag" to try break it) and OpenBox based WM, where Alt-Tab and clicking the task-bar to maximize would previously not repaint as required. All working as expected now.

PR #465, opened 2024-05-11
Don't override a user request for a specific nightly version

bjorn3:
Without this change running makepad with a locally build version of rustc_codegen_cranelift doesn't work as rustc_codegen_cranelift needs a particular nightly rustc version to pair with and trying to load it with any other version will crash the compiler. cargo-clif handles using the right nightly version for you, but makepad overrides it before this change, causing the aformentioned crash.

### DIFF
--- a/studio/src/build_manager/build_server.rs
+++ b/studio/src/build_manager/build_server.rs
@@ -16,6 +16,7 @@ use {
     },
     std::{
         collections::HashMap,
+        env,
         fmt,
         path::PathBuf,
         sync::{Arc, RwLock, Mutex, mpsc::Sender},
@@ -92,9 +93,6 @@ impl BuildConnection {
         let args: Vec<String> = match &what.target {
             BuildTarget::ReleaseStudio => vec![
                 "run".into(),
-                "nightly".into(),
-                "cargo".into(),
-                "run".into(),
                 "-p".into(),
                 what.binary.clone(),
                 "--message-format=json".into(),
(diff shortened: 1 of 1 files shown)


bjorn3:
Locally (x86_64 Linux) building makepad-studio from scratch in debug mode takes 15s with the default LLVM backend and 10s with rustc_codegen_cranelift. For several trivial changes I tried it speeds up compile time for makepad-studio anywhere between 20% and 70% depending on what I change. I haven't tried changing any of the other projects as the embedding of programs in the editor is currently broken on Linux due to a missing WebSockets implementation. Be aware that rustc_codegen_cranelift is currently broken on arm64 macOS due to an abi issue (https://github.com/rust-lang/rustc_codegen_cranelift/issues/1248). It should work under Rosetta2, but the overhead of that is likely to make it a net perf loss.

studio/src/build_manager/build_server.rs
```diff
@@ -281,15 +236,21 @@ impl BuildConnection {
...
             ("MAKEPAD_STUDIO_HTTP", http.as_str()),
             ("MAKEPAD", "lines")
         ];
 
-        let process = ChildProcess::start("rustup", &args, path, &env).expect("Cannot start process");
-        
+        // Default to nightly rustc but don't overwrite any user request for a
+        // specific nightly version.
+        if !env::var("RUSTUP_TOOLCHAIN").map_or(false, |toolchain| toolchain.contains("nightly")) {
+            env.push(("RUSTUP_TOOLCHAIN", "nightly"));
```
bjorn3, line 252: The `RUSTUP_TOOLCHAIN` env var is read by rustup to determine which toolchain to use when the `+toolchain` cli arg or `rustup run` isn't used and set for any command it invokes with the toolchain that is actually used.

PR #466, opened 2024-05-13
[Ergonomics] Add `has_texture` function to `Image`.

noxware:
Allow knowing if an image has an image set or not.

As a workaround to do the same, you can currently use `size_in_pixels` and see if it returns `None` knowing that that happens when there is no texture set, but that's an indirect way to ask and requires the context parameter that is not necessary for knowing if there is a texture or not.

The `has_texture` function added allow us to directly ask that, without exposing internals.

- [Example without `has_texture`](https://github.com/noxware/makepad/blob/6660a029f8915db9056f4afe340687b5b99475ca/examples/image_viewer/src/ui.rs#L50)

- [Example with `has_texture`](https://github.com/noxware/makepad/blob/8200bb07751a43bc709457cf6108032c9434b0c4/examples/image_viewer/src/ui.rs#L49)

### DIFF
--- a/widgets/src/image.rs
+++ b/widgets/src/image.rs
@@ -56,6 +56,11 @@ impl Image {
             .and_then(|t| t.get_format(cx).vec_width_height())
     }
 
+    /// True if a texture has been set on this `Image`.
+    pub fn has_texture(&self) -> bool {
+        self.texture.is_some()
+    }
+
     pub fn draw_walk(&mut self, cx: &mut Cx2d, mut walk: Walk) -> DrawStep {
         // alright we get a walk. depending on our aspect ratio
         // we change either nothing, or width or height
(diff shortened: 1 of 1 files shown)


widgets/src/image.rs
```diff
@@ -168,5 +173,14 @@ impl ImageRef {
             None
         }
     }
+
+    /// See [`Image::has_texture()`].
+    pub fn has_texture(&self) -> bool {
+        if let Some(inner) = self.borrow() {
+            inner.has_texture()
+        } else {
+            false
```
noxware, line 182: When the internal ref cell is already mutably borrowed and this fails the options are:
- Return `false` (current behavior).
- Return Option or Result (exposing the fact that this is internally a ref cell).
- Simply blow (panic/unwrap/expect). This also exposes the fact that there is a ref cell involved but lets the user dev know about a borrowing issue (at runtime). So I would prefer this but it would not be consistent with the silent fails behavior in the rest of the widget.

None of the options are ideal, but they are the only options I can think of if there is a RefCell behind the scenes.

PR #485, opened 2024-06-28
consistent meaning of value at slider widget + set_value

noxware:
# Problem
- Given a variable `slider` of type `SliderRef`, if you do `slider.value()` you get the current value (acording to what you defined as possible values in the slider configuration).
- But, if you do `slider.borrow().value` you get something different. You get the relative position of the slider (from 0 to 1).

This is confusing, and also makes hard to set a value programmatically for the slider as you would need to do `slider.borrow_mut().value = /* insert calculations based on the min/max configuration to translate to the relative value */`. And there are no getters for `min` and `max` anyways. 

> Please note that is not only the `value()` method, but also the `slided()` event returns the absolute value, not the relative one.

# Changes

- Adds a `value()` function in the `Slider` impl.
- Make the existing `value()` function at `SliderRef` just call that.
- Adds `set_value` functions to both struct impls.
  - This is to support what I was originally trying [here](https://github.com/moxin-org/moxin/blob/7dd935d2a927b2aa8833755ed2ad3d4dac016591/src/chat/chat_params.rs#L175).
- Changes the name of the `value` field of the `Slider` struct, so it's not just "value" anymore.
  - This is technically a **breaking change** as this field is public. Although I don't see is being used in code of the makepad repo.

### DIFF
--- a/widgets/src/slider.rs
+++ b/widgets/src/slider.rs
@@ -133,8 +133,11 @@ impl Slider {
     }
 
     pub fn set_value(&mut self, v: f64) {
+        let prev_value = self.value();
         self.set_internal(v);
-        self.update_text_input();
+        if v != prev_value {
+            self.update_text_input();
+        }
     }
 }
 
--- a/widgets/src/slider.rs
+++ b/widgets/src/slider.rs
@@ -62,7 +62,7 @@ pub struct Slider {
     
     #[live] bind: String,
     
-    #[rust] pub value: f64,
+    #[rust] pub relative_value: f64,
     #[rust] pub dragging: Option<f64>,
 }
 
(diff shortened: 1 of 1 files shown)


widgets/src/slider.rs
```diff
@@ -127,6 +127,18 @@ impl Slider {
...
     }
+
+    pub fn value(&self) -> f64 {
+        self.to_external()
+    }
+
+    pub fn set_value(&mut self, v: f64) {
+        let prev_value = self.value();
+        self.set_internal(v);
+        if v != prev_value {
```
noxware, line 138: If you are binding the value of this slider to something external, by calling `set_value`, then, we should not touch the text input unless the value being set is actually different.

This is because would disallow you to have a dirty/incomplete state in the text input, which means you may not be abale to edit the slider using the text input.

I'm pretty sure this would work as

```rust
pub fn set_value(&mut self, v: f64) {
    if v != self.value() {
        self.set_internal(v);
        self.update_text_input();
    }
}
```

But since most widgets always set their internal state unconditionally, I would like to avoid not matching that behavior here. So calling always `set_internal` and just focusing on not touching the text input.

PR #488, opened 2024-07-06
Demo of infinite loop in ShapeCache: run `html_experiment` to crash Makepad

kevinaboos:
DO NOT MERGE. This is just an example.

The failure occurs when trying to draw emoji. See discord for more.


kevinaboos:
This has been fixed as well.

PR #497, opened 2024-07-23
Don't require a mutable borrow for `find_widgets()`

kevinaboos:
This function previously required a mutable borrow of `self` (borrowing the widget instance itself) in order to traverse a path and return a reference to a particular widget.
This was only done to update the cache of found widgets, in order to accelerate future searches for the same path ID.

This is not necessary, and in fact it results in Makepad app devs having to adopt unnecessary clones and other inefficient code patterns in more complex app code.

This changeset enables code like this to compile:
```rust
fn handle_event(...) {
    // this is an immutable borrow of `self`. The `info` state is massive.
    let Some(info) = self.info.as_ref() else { return };

    if let Event::Actions(actions) = event {
        // this use to fail due to *mutably* borrowing `self`.
        if self.button(id!(copy_link_to_user_button)).clicked(actions) {
             .... 
        }

        if let Some(room_member) = info.room_member.as_ref() {
            if self.button(id!(ignore_user_button)).clicked(actions) {
                ....
            }
        }

        // here, there are a bunch more path searches similar to `self.button(...)`
    }
}
```

### DIFF
--- a/code_editor/src/code_editor.rs
+++ b/code_editor/src/code_editor.rs
@@ -323,7 +323,7 @@ impl CodeEditor {
         self.walk
     }
     
-    pub fn find_widgets(&mut self, _path: &[LiveId], _cached: WidgetCache, _results: &mut WidgetSet){
+    pub fn find_widgets(&self, _path: &[LiveId], _cached: WidgetCache, _results: &mut WidgetSet){
     }
         
     pub fn draw_walk_editor(&mut self, cx: &mut Cx2d, session: &mut Session, walk:Walk) {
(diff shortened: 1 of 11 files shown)


jmbejar:
@kevinaboos Just built and run Moxin against this PR and nothing seems broken. I only had to change in a few custom widgets where we were implementing our own version of `find_widgets`, but no a big deal.

I haven't been looking for opportunities to improve code where we are now allowed to do so (because of this change), just wanted to double check this change is safe enough (although it is not fully backward compatible).

kevinaboos:
Nice, glad to hear it. I was able to make a few additional improvements to Robrix by looking for:
1. places where I queried IDs (by searching for `id!(`)
2. How `self` was borrowed, primarily in `handle_event()` and `draw_walk()`
3. Invocations of `clone()`, i.e., any unnecessary clones

PR #504, opened 2024-07-31
Bring moxin toggle panel

noxware:
Screenshot 2024-07-31 at 12 17 14

Screenshot 2024-07-31 at 12 17 24


widgets/src/base.rs
```diff
@@ -973,6 +974,12 @@ live_design!{
     BareStep = <BareStep>{}
     TurtleStep = <TurtleStep>{}
     ColorPicker = <ColorPicker>{}
+
+    // Re-export the icons in case you want to override the built-in buttons with
```
noxware, line 978: Example of this at `theme_desktop_dark.rs` and at https://github.com/moxin-org/moxin/blob/30431819284744ad23179562f66fea86781485b6/src/shared/widgets.rs#L487

PR #509, opened 2024-08-16
PortalList improvements: sanity check for first ID within item range

kevinaboos:
This emits a warning (on debug builds only) when the user has made a mistake using PortalList -- specifically, when the user has forgotten to correctly set either the first item ID or the item range to be displayed by the list.

The reason for this sanity check is that it is almost impossible to otherwise debug what the problem is when the PortalList is showing a completely blank view. Personally, I experienced this frequently in Robrix when the underlying item set changes, and either the item range or first ID was improperly updated to reflect the new items list. Hopefully this will save future devs from the same headache.

Also, this PR adds the `PortalList::is_at_end()` method to help users determine whether the list is showing the last item in the item range.

It also adds documentation to several methods, which should help clarify some things. @makepaddev kindly check the docs I added for `set_first_id_and_scroll()` and `scroll_position()`, I'm not 100% sure if they're fully correct. I wanted to add that because some external open-source devs have misunderstood and thus improperly used the result of `scroll_position()`.

### DIFF
--- a/widgets/src/portal_list.rs
+++ b/widgets/src/portal_list.rs
@@ -307,6 +307,17 @@ impl PortalList {
         if let Some(draw_state) = self.draw_state.get() {
             match draw_state {
                 ListDrawState::Begin => {
+                    // Sanity check: warn on the first item ID being outside of the previously-set item range.
+                    // This check is done here rather than in `begin()`, as most PortalList usage doesn't set
+                    // the item range properly until right before looping over `next_visible_items()`.
+                    #[cfg(debug_assertions)]
+                    if self.fails_sanity_check_first_id_within_item_range() {
+                        warning!("PortalList: first_id {} is greater than range_end {}.\n\
+                            --> Check that you have set the correct item range and first item ID!",
+                            self.first_id, self.range_end,
+                        );
+                    }
+
                     let viewport = cx.turtle().padded_rect();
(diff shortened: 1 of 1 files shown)


widgets/src/portal_list.rs
```diff
@@ -791,19 +816,27 @@ impl Widget for PortalList {
 }
 
 impl PortalListRef {
+    /// Sets the first item to be shown and its scroll offset.
+    ///
+    /// On the next draw pass, this PortalList will draw the item with the given `id`
+    /// as the first item in the list, and will set the *scroll offset*
+    /// (from the top of the viewport to the beginning of the first item)
```
kevinaboos, line 823: Please double-check this comment about what `s` represents.

widgets/src/portal_list.rs
```diff
@@ -813,37 +846,52 @@ impl PortalListRef {
...
+
+    /// Returns whether the given `actions` contain an action indicating that this PortalList was scrolled.
     pub fn scrolled(&self, actions: &Actions) -> bool {
         if let PortalListAction::Scroll = actions.find_widget_action(self.widget_uid()).cast() {
             return true;
         }
         false
     }
 
+    /// Returns the current scroll offset of this PortalList.
```
kevinaboos, line 874: Please double-check this comment -- is "scroll offset" a good term for this?

PR #510, opened 2024-08-16
PortalList: add method that returns each visible item's scroll offset with its item id

kevinaboos:
New method: `PortalList::next_visible_item_with_scroll()` is similar to `next_visible_item()`, but in addition to the next visible item's index/ID, it also returns the item's scroll offset from the top of the `PortalList`'s viewport.

This makes it possible for a user of `PortalList` to track where each visible item has been drawn, which is useful for myriad reasons.


makepaddev:
Seems fine, if it works for you its good. Maybe we should make it another method to not break existing code

PR #525, opened 2024-09-02
Solved the issue of panic caused by special characters, and added a font library 

ZhangHanDong:
Solved the issue of panic caused by special characters, and added a font library containing special characters and an open-source commercial Chinese font library.

**Problem description:**

When entering a character like `ﭿ`, a panic occurred.

```rust
thread 'main' panicked at draw/src/font_atlas.rs:129:26:
begin <= end (3 <= 0) when slicing `ﭿ `
```

This error indicates a problem occurred when processing text slices. The error message shows `start=3, end=0, text.len()=4`, which means the end index is less than the start index, making it an invalid range.

**The pr modification mainly made the following improvements:**

1. `font_atlas.rs` : 

- Ensuring that end is always greater than start.
- Added additional boundary checks.
- Using end.min(text.len()) to prevent out-of-bounds errors.
- Added checks for empty text slices.

These modifications should be able to handle special characters like ﮝ and prevent invalid text slices.

Modify `makepad-example-simple` to use the special character font library in TextInput, and test special characters, verify the issue has been fixed.

Screenshot 2024-09-03 at 07 19 37

2. Add font library (`DejaVuSans.ttf`) for displaying special characters, such as "⚙ / � / 𝌆 / 𝍖 " ...

3. Newly added open source free commercial Chinese font library.  [LXGW WenKai (霞鹜文楷)](https://github.com/lxgw/LxgwWenKai)  `LXGWWenKaiBold.ttf` / `LXGWWenKaiLight.ttf` / `LXGWWenKaiMonoLight.ttf` /  `LXGWWenKaiRegular.ttf` .  

4. Add `makepad-example-simple-fonts` to demonstrate how to use the newly added font library. 

simple-fonts


makepaddev:
We added the fonts separately, should work now as well

makepaddev:
should be fixed alread

PR #532, opened 2024-09-09
Add demo of incorrect text wrapping to text-flow example

kevinaboos:


kevinaboos:
Demo video here: https://drive.google.com/file/d/12wNVQWf_hdSUaOEZDcXmWp6EGqdjBSeY/view?usp=sharing

ejpbruel2:
This is a regression from a fix I landed yesterday. The layout algorithm accidentally reports soft line breaks to the turtle as hard line breaks, and this messes up the turtle's own word wrap algorithm. It should be fixed now in the latest rik branch

kevinaboos:
Thanks Eddy, can confirm that it appears to be fixed now!

PR #543, opened 2024-09-14
add trace using perfetto trace format

hypheng:
add some basic trace to some main process, only for android now.
Pull the trace file from android directory like </storage/emulated/0/Android/data/${APP_PACKAGE}/files/makepad.ptrace>
Use https://ui.perfetto.dev to display the trace file.


makepaddev:
Not going to merge adding this dependency here, also its already conflicting at this point.

PR #545, opened 2024-09-19
Add `filter_widget_actions` to `WidgetActionApi`

syf20020816:


univerz:
(i'm not a maintainer, but) why not return an iterator to remove Vec allocation?

```rust
    fn filter_widget_actions(&self, widget_uid: WidgetUid) -> impl Iterator<Item = &WidgetAction> {
        self.iter().filter_map(move |action| {
            action
                .downcast_ref::<WidgetAction>()
                .and_then(|action| (action.widget_uid == widget_uid).then_some(action))
        })
    }
```

makepaddev:
Hmm i like the iterator idea. Didnt know you could write it like that. The alloc of the Vec is extremely rare tho and at a scale of '1x per user click', so its an optimisation that will do very little.

syf20020816:
I optimize the function and add the doc
I accidentally formatted the code, which resulted in many lines of changes, but in reality, only the following code was modified
```rust
    /// ## Filter all actions by widget uid
    /// this function use to filter all actions from `Event::Actions(actions)`,
    /// if multi actions in same widget may happened in the same time, this function will help you get all
    /// and back an Iter
    /// ### Attention
    /// **If you want to focus on target actions and need to cast directly use `filter_widget_action_cast`**
    /// ### Examples
    /// #### find and directly do target action without param
    /// you can `filter_widget_actions` and then do find to get target action you want,
    /// then do map to do want you what
    /// ```rust
    /// let actions = cx.capture_actions(|cx| self.super_widget.handle_event(cx, event, scope));
    ///
    /// self.gbutton(id!(auto_connect)).borrow().map(|x| {
    ///     let mut actions = actions.filter_widget_actions(x.widget_uid());
    ///     actions.find(|action| {
    ///         if let GButtonEvent::Clicked(_) = action.cast(){
    ///             true
    ///         }else{
    ///             false
    ///         }
    ///     }).map(|action|{
    ///         dbg!(action);
    ///     });
    /// });
    /// ```
    /// #### find and cast
    /// ```rust
    /// let actions = cx.capture_actions(|cx| self.super_widget.handle_event(cx, event, scope));
    ///
    /// self.gbutton(id!(auto_connect)).borrow().map(|x| {
    /// let actions = actions.filter_widget_actions(x.widget_uid());
    ///     actions.for_each(|action| {
    ///         if let GButtonEvent::Clicked(param) = action.cast(){
    ///             dbg!(param);
    ///         }
    ///     })
    /// });
    /// ```
    fn filter_widget_actions(&self, widget_uid: WidgetUid) -> impl Iterator<Item = &WidgetAction>;
    /// ## Filter widget actions by widget id and cast
    /// this function can help you cast the widget actions to the widget you want, the diff is:
    /// - try cast all widget actions (This method is not recommended when a large number of actions occur simultaneously)
    /// - back `Iterator<Item = T>` not `Iterator<Item = &T>`
    /// ### Example
    /// ```rust
    /// self.gbutton(id!(auto_connect)).borrow().map(|x| {
    /// let actions = actions.filter_widget_actions_cast::<GButtonEvent>(x.widget_uid());
    ///     actions.for_each(|action| {
    ///         if let GButtonEvent::Clicked(param) = action{
    ///             dbg!(param);
    ///         }
    ///     })
    /// });
    /// ```
    fn filter_widget_actions_cast<T: WidgetActionTrait + 'static + Send>(
        &self,
        widget_uid: WidgetUid,
    ) -> impl Iterator<Item = T>
    where
        T: Default + Clone;
```

PR #554, opened 2024-10-03
(poc) wasm-bindgen support

univerz:
warning: i don't know js or wasm internals => especially the changes in makepad's js files (necessary for threading) are completely above my pay grade, but now things like chrono (& makepad's websockets) work on my machine.

i made some fixes in the first 2 commits, the relevant changes are in the last one.

### DIFF
--- a/platform/src/os/web/web.js
+++ b/platform/src/os/web/web.js
@@ -451,21 +451,27 @@ export class WasmWebBrowser extends WasmBridge {
     }
     
     alloc_thread_stack(context_ptr, timer) {
-        let tls_size = this.exports.__tls_size.value;
-        tls_size += 8 - (tls_size & 7); // align it to 8 bytes
-        let stack_size = this.thread_stack_size; // 8mb
-        if ((tls_size + stack_size) & 7 != 0) throw new Error("stack size not 8 byte aligned");
-        let tls_ptr = this.exports.wasm_thread_alloc_tls_and_stack((tls_size + stack_size) >> 3);
-        this.update_array_buffer_refs();
-        let stack_ptr = tls_ptr + tls_size + stack_size - 8;
-        return {
-            tls_ptr,
-            stack_ptr,
+        var ret = {
             timer,
             module: this.wasm._module,
             memory: this.wasm._memory,
             context_ptr
+        };
+        if (typeof this.exports.__wbindgen_start !== 'undefined') {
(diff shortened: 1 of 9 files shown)


platform/src/os/web/web.js
```diff
@@ -451,21 +451,27 @@ export class WasmWebBrowser extends WasmBridge {
     }
     
     alloc_thread_stack(context_ptr, timer) {
-        let tls_size = this.exports.__tls_size.value;
```
univerz, line 454: the problem is that wasm-bindgen removes [`__tls_size`](https://github.com/rustwasm/wasm-bindgen/blob/4765448545d7bfc0223fd52f25d966b0bfbf8004/crates/threads-xform/src/lib.rs#L140)

tools/cargo_makepad/src/wasm/compile.rs
```diff
@@ -151,7 +175,16 @@ pub fn build(config:WasmConfig, args: &[String]) -> Result<WasmBuildResult, Stri
             
             cp_brotli(&dep_dir.join("src/os/web/web_gl.js"), &app_dir.join("makepad_platform/web_gl.js"), false, config.brotli)?;
             
-            cp_brotli(&dep_dir.join("src/os/web/web_worker.js"), &app_dir.join("makepad_platform/web_worker.js"), false, config.brotli)?;
+            if config.bindgen {
+                let jsfile = dep_dir.join("src/os/web/web_worker.js");
+                let js = std::fs::read_to_string(&jsfile).map_err(|e| format!("Unable to find web.js {e:?}"))?;
+                let tmp = build_dir.join("web_worker.js");
+                let mut file = std::fs::OpenOptions::new().write(true).truncate(true).create(true).open(&tmp ).unwrap();
+                file.write(format!("import init from '../bindgen.js';\n{js}").as_bytes()).unwrap();
```
univerz, line 179: in the first version i used dynamic import, but it probably wouldn't work well with html minification. is there a better way?

univerz:
btw @makepaddev could we make this optional, for example behind (`cargo-makepad wasm`) `--watch`?

```<script type='module' src='./makepad_platform/auto_reload.js'></script>```

univerz:
and final notification to your inbox to clean the logs completely:

* what would be the best way to fix missing `favicon.ico`?

* firefox console warning:
```WEBGL_debug_renderer_info is deprecated in Firefox and will be removed. Please use RENDERER.```

PR #564, opened 2024-10-18
fix label widget text empty lazy draw bug

syf20020816:
If set `<Label>{text: "",}` , `set_text_and_redraw` redraw the text and show the new text, seems need to set text a space when the label first draw_walk

### DIFF
--- a/widgets/src/label.rs
+++ b/widgets/src/label.rs
@@ -40,6 +40,11 @@ impl Widget for Label {
     fn draw_walk(&mut self, cx: &mut Cx2d, _scope: &mut Scope, walk:Walk)->DrawStep{
         let walk = walk.with_add_padding(self.padding);
         cx.begin_turtle(walk, Layout::default());
+        // here we need to check if the text is empty, if so we need to set it to a space
+        // or the text draw will not work(seems like lazy drawtext bug)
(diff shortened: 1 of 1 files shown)


univerz:
@eddybruel i think this should be solved at `draw_text` level - `text_input` has the same problem.

makepaddev:
I think i need to change the 'and_redraw' stuff to be 'check if in redrawflow, IF NOT, trigger redraw'. this explicit redraw is causing problems.

univerz:
not sure if we talk about the same thing. this is how the input inside the RoundedView looks like with:
* `set_text`
* `empty_message: ""` <- this is case where draw_text doesn't "set" height for empty text
* without `empty_message` (note the ugly default `0` from theme_desktop_dark, probably set because of this bug)

image

univerz:
(i assumed that this PR renders a fake space in the empty case for the same reason)

syf20020816:
Fake space in label aim to solve if is empty at beginning then use `set_text_and_redraw` still can not display the new text, 
This is indeed a known problem. This fix may not the final solution.
And the input is not yet perfect, use `0` as a default is not a good way. We should add it into todo list (I think use `placeholder` and bind a default height will be better just like `HTML`)

PR #567, opened 2024-10-23
Fix live-reloading in Linux

yulnr:


univerz:
it would be great to have this disabled by default

yulnr:
@univerz what's does live-reloading do that you'd want it disabled? I'm guessing state changes happen after updates that you don't want? Or are you talking about not having the file watcher running on release builds, or just simply not run unless explicitly requested etc.? (resource usage)

univerz:
> Or are you talking about not having the file watcher running on release builds, or just simply not run unless explicitly requested etc.? (resource usage)

this; it spawns a thread in the startup "hotpath" + tries to load each file every 100ms. it also makes sense outside of release builds - the design phase is <10% of total dev time => from my point of view using env var / parameter seems like a good tradeoff in this case, even if it affects discoverability (we should collect all these things in the readme anyway).

yulnr:
> > Or are you talking about not having the file watcher running on release builds, or just simply not run unless explicitly requested etc.? (resource usage)
> 
> this; it spawns a thread in the startup "hotpath" + tries to load each file every 100ms. it also makes sense outside of release builds - the design phase is <10% of total dev time => from my point of view using env var / parameter seems like a good tradeoff in this case, even if it affects discoverability (we should collect all these things in the readme anyway).

Agree, I'll discuss with Rik. (we have a similar scenarios like benchmarking on mobile while having unnecessary websocket connections going on for debugging purposes).
I also don't mind env var / parameters. But for what I've been doing in the last year in Makepad and across applications, I'm using the live-reloading at least like a good 60% of the time. I'd say that being able to turn it off is very important, but wether the default should be on or off is arguable (e.g. Flutter, SwiftUI, React Native, etc. have it on by default on debug builds).

(In this PR I'm just matching the behaviour from other platforms)

PR #574, opened 2024-10-31
Derive more traits on `WidgetUid` to make it usable in Hashed data structures

kevinaboos:
like HashMap, etc.


kevinaboos:
updated to latest `rik` HEAD

PR #589, opened 2024-11-12
TextInput: expose current keyboard modifiers in `Return` action

kevinaboos:
This is necessary for applications that wish to handle the `Return` key being pressed in different ways, based on the active modifier keys.

Also adds new `returned_with_modifiers()` function to make it easier for users to query this info from the `TextInput` widget.


kevinaboos:
Hmm, actually I suppose this can already be achieved with `key_down_unhandled()`,  but I think that API itself ought to be renamed/redesigned a bit to clarify under which conditions (key presses) this action will actually be emitted.

PR #599, opened 2024-11-19
Make `PortalList::smooth_scroll_to()` only include a specified maximum number of items in its scrolling animation

kevinaboos:
This maximum number of items can be set via an optional parameter; the default is 20 items maximum.

Also, simplify and document the APIs for `smooth_scroll_to()` and `smooth_scroll_to_end()`


kevinaboos:
Closed in favor of #600

PR #609, opened 2024-12-03
`Cursor::Grab` and `Cursor::Grabbing` on different platform

Guocork:


Guocork:
I couldn't find the matching cursor name on Windows. I tested it on Linux and the web, and it works fine.

yulnr:
@Guocork thanks for working on this.
The changes are correct for the platforms you mentioned. When it comes to Windows, it does not support these type of cursors which would require us to include our own custom cursor files. 
We can skip that step for now (until we discuss that implementation and what it involves), but we need to update the Windows platform code to handle the new cursor types, otherwise it does not compile due to not matching on all variants of the enum:
in `win32_app.rs`
```rust
    pub fn set_mouse_cursor(&mut self, cursor: MouseCursor) {
        if self.current_cursor.is_none() || self.current_cursor.unwrap() != cursor {
            let win32_cursor = match cursor {
                MouseCursor::Hidden => {
                    PCWSTR::null()
                },
                MouseCursor::Default => IDC_ARROW,
                MouseCursor::Crosshair => IDC_CROSS,
                MouseCursor::Hand => IDC_HAND,
                // Default to Hand for non-supported cursors, until we include our own custom cursor files. (<-- Add this comment)
                MouseCursor::Grab | MouseCursor::Grabbing => IDC_HAND,  // (<-- add this line)
```
That should fix it for now.

Guocork:
> @Guocork thanks for working on this. The changes are correct for the platforms you mentioned. When it comes to Windows, it does not support these type of cursors which would require us to include our own custom cursor files. We can skip that step for now (until we discuss that implementation and what it involves), but we need to update the Windows platform code to handle the new cursor types, otherwise it does not compile due to not matching on all variants of the enum: in `win32_app.rs`
> 
> ```rust
>     pub fn set_mouse_cursor(&mut self, cursor: MouseCursor) {
>         if self.current_cursor.is_none() || self.current_cursor.unwrap() != cursor {
>             let win32_cursor = match cursor {
>                 MouseCursor::Hidden => {
>                     PCWSTR::null()
>                 },
>                 MouseCursor::Default => IDC_ARROW,
>                 MouseCursor::Crosshair => IDC_CROSS,
>                 MouseCursor::Hand => IDC_HAND,
>                 // Default to Hand for non-supported cursors, until we include our own custom cursor files. (<-- Add this comment)
>                 MouseCursor::Grab | MouseCursor::Grabbing => IDC_HAND,  // (<-- add this line)
> ```
> 
> That should fix it for now.

Ok, I fix that.

yulnr (approved):

PR #611, opened 2024-12-06
Slider enhancement

Guocork:


widgets/src/slider.rs
```diff
@@ -475,14 +475,23 @@ impl Widget for Slider {
...
                 self.animator_play(cx, id!(hover.on));
-            }
+            },
             Hit::FingerHoverOut(_) => {
                 self.animator_play(cx, id!(hover.off));
             },
-            Hit::FingerDown(_fe) => {
+            Hit::FingerHoverOver(_) => {
+                cx.set_cursor(MouseCursor::Hand);
+            },
```
yulnr: I believe this should be instead:
```rs
Hit::FingerHoverOver(_) => {
    cx.set_cursor(MouseCursor::Grab);
},
```
Because we want the open hand to show as soon as we hover on the slider (instead of the pointing finger of MouseCursor::Hand).

On a side-note, we should actually use `MouseCursor::Hand` if the hover is not on top of the slider knob (if the cursor is on top of the slider line but not on top of the circle). But I don't think this easily detectable, so maybe these changes are good enough for now.

yulnr (commented):
Great job! Leaving a comment about cursor usage

widgets/src/slider.rs
Guocork: Yeah, with the current component, it's hard to calculate the area of the slider knob, so I went with the current solution.

PR #612, opened 2024-12-12
Fixed Bug: Index error when handling multi-byte characters (such as Chinese question mark)

ZhangHanDong:
This issue was discovered in the moly app, where the app crashes when the last character in the text box is a Chinese question mark.

**Mainly modify the code corresponding to line 664.** There are some format interferences in the commit, which can be ignored.

For a detailed description, see the issue : [moly issue 326](https://github.com/moxin-org/moly/issues/326).

This fix has made the following improvements:

- Improved the calculation of glyph_end:
   - For the last glyph, use string.len() instead of the cluster value of the next glyph. 
   - This ensures that we always have a valid end position.
- Added safety checks:
    - Ensure that glyph_end is greater than glyph_start. 
    - Check for exceeding text boundaries. 
    - Use the min function to ensure that the text range is not exceeded.
- Special handling of boundary conditions:
    - When processing the last glyph, use a safer boundary calculation method. 
    - Avoided index errors when handling multi-byte characters (such as Chinese question mark).

With these modifications, the bug has been fixed.

### DIFF
--- a/draw/src/font_atlas.rs
+++ b/draw/src/font_atlas.rs
@@ -48,14 +48,14 @@ pub struct CxFontAtlas {
 }
 
 /// A cache for rasterized glyph data.
-/// 
+///
 /// The cache is split into two parts:
 /// - The rasterized glyph data itself.
 /// - An index for the rasterized glyph data.
-/// 
+///
 /// The index contains a mapping from the `LiveId` of a glyph to an index entry. Each index entry
 /// contains the offset of the glyph within the rasterized glyph data, as well as its size.
-/// 
+///
 /// This cache can optionally be backed by a file on disk, allowing the cache to be persisted across
 /// runs of the application. This should improve startup time, as the cache data does not need to be
 /// re-generated every time the application is started.
@@ -73,12 +73,12 @@ pub struct FontCache {
 
 impl FontCache {
     /// Creates a new `FontCache`.
-    /// 
+    ///
     /// If a directory is provided, the cache data will be initialized from the file `font_cache`,
     /// and the cache index from the file `font_cache_index`. Moreover, every time a new glyph is
     /// inserted into the cache, the data for the glyph will be appended to `font_cache`, and the
     /// index entry for the glyph to `font_cache_index`.
-    /// 
+    ///
     /// If no directory is provided, the cache will start out empty, and will not be backed by a
     /// file on disk.
     fn new(dir: Option<impl AsRef<Path>>) -> Self {
@@ -168,7 +168,7 @@ impl FontCache {
             buffer[8..16].copy_from_slice(&u64::try_from(offset).unwrap().to_be_bytes());
             buffer[16..24].copy_from_slice(&u64::try_from(size.width).unwrap().to_be_bytes());
             buffer[24..32].copy_from_slice(&u64::try_from(size.height).unwrap().to_be_bytes());
-            
+
             index_file
                 .write_all(&buffer)
                 .expect("couldn't write to font cache index file");
@@ -238,7 +238,7 @@ impl CxShapeCache {
             let font = font_atlas.fonts[font_id].as_mut().unwrap();
             let glyph_id = font.glyph_id('•').0 as usize;
             if glyph_id == 0 {
-                None 
+                None
             } else {
                 Some((font_id, glyph_id))
             }
@@ -269,6 +269,7 @@ impl CxShapeCache {
                 font_ids,
                 font_atlas,
                 &mut glyph_infos,
+                0,
             );
             if self.shape_keys.len() == 4096 {
                 let shape_key = self.shape_keys.pop_front().unwrap();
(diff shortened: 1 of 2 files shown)


draw/src/shader/draw_text.rs
```diff
@@ -661,9 +661,21 @@ impl DrawText {
                         let mut iter = glyph_infos.iter().peekable();
                         while let Some(glyph_info) = iter.next() {
                             let glyph_start = start + glyph_info.cluster;
-                            let glyph_end = start + iter.peek().map_or(string.len(), |glyph_info| glyph_info.cluster);
+
+                            let glyph_end = if let Some(next_info) = iter.peek() {
```
ejpbruel2: I'm confused about this change. Doesn't this do exactly the same as the original line?

draw/src/shader/draw_text.rs
```diff
@@ -661,9 +661,21 @@ impl DrawText {
...
+                                start + next_info.cluster
+                            } else {
+                                // If it is the last glyph, use the length of the entire string
+                                start + string.len()
+                            };
+                            // Add safety checks.
+                            if glyph_end <= glyph_start || glyph_start >= text.len() {
```
ejpbruel2: This safety check should not be necessary.

The following assumptions should hold about cluster values:
- They are strictly increasing
- They should always be less than or equal to the length of the text

Given these assumptions, since glyph_start is defined as start + glyph_info.cluster, and glyph_end as either start + next_glyph_info.cluster, or the length of the text, then glyph_start < glyph_end should always hold.

Apparently, this is not the case, or you wouldn't have added this check, so one or both of the above assumptions must be broken. That's a bug, and we should fix it accordingly, rather than add special casing to deal with the bug here.

draw/src/shader/draw_text.rs
```diff
@@ -661,9 +661,21 @@ impl DrawText {
...
+                                continue;
+                            }
+                            // Use safe maximum boundaries
+                            let safe_end = glyph_end.min(text.len());
```
ejpbruel2: This check also should not be necessary, for the same reasons as outlined above.

ejpbruel2 (changes requested):
I understand what this commit is trying to do, and although it technically solves the problem, I believe the solution is wrong, in the sense that it adds special handling to work around the actual bug. That just increases complexity while leaving the actual bug in place.

The actual bug is that this code was written with the following assumptions in mind:
- Cluster values are strictly increasing
- Cluster values should always be less than the length of the text

Apparently, these assumptions do not hold in practice (or you would not have written this code). What we should do instead is look at the code that computes the cluster values and check why it breaks these assumptions.

ZhangHanDong:
@ejpbruel2 Thank you for the feedback. I am still a novice in font rendering, and I will do my best to correct and solve the problem according to your feedback.

ZhangHanDong:
# Analysis of Cluster Value Calculation in Makepad Text Rendering System

## Background

- In the Makepad text rendering system, text processing first goes through makepad_rustybuzz, which is a Rust implementation of the HarfBuzz algorithm. This layer is responsible for converting Unicode text into glyph sequences, a fundamental step in text rendering.
- When text passes through makepad_rustybuzz for processing, it generates cluster values. These values are designed to reflect the positional relationships between characters in the text - essentially acting as positioning markers for where each character should be rendered. However, we observed an anomaly in these cluster values specifically when processing Chinese question marks, where the expected positional relationships were not maintained.

## Problem Description

The system crashes when processing text ending with Chinese punctuation marks. For example, when processing the text "什么是 Mofa ？", the system reports an error:

```
begin <= end (14 <= 10) when slicing `什么是 Mofa ？`
```

Interestingly, this issue only occurs when text ends with Chinese punctuation marks, while the system works normally when processing text ending with regular Chinese characters.

## In-depth Analysis

### Normal Processing Flow

When the system processes regular Chinese text, text processing works like an orderly production line:

```
Text: 你好世界
Byte positions: [你]0-2  [好]3-5  [世]6-8  [界]9-11
Cluster values:    0       3        6        9
```
In this case, each character occupies a fixed three bytes, and cluster values increase regularly according to byte positions, allowing the system to accurately locate each character's position.

### Problem Trigger Scenario

When text ends with Chinese punctuation marks, the processing flow changes:

1. The system first processes text using the primary font
2. Upon encountering Chinese punctuation, it finds no corresponding glyph in the primary font
3. Font fallback mechanism is triggered, attempting to use backup fonts
4. Backup font generates new glyph information and cluster values
5. System continues processing text, but the continuity of cluster values is broken

This is like suddenly needing to use a special workstation with a different numbering system on a production line, leading to confusion in the subsequent assembly process.

## Root Cause Summary
The root cause of this issue can be attributed to three interconnected factors:

1. Font Fallback Trigger: Chinese punctuation (like "？") requires special font support, triggering the font fallback mechanism
2. Cluster Value Calculation Change: Backup fonts may use different cluster calculation rules during processing
3. Processing Continuity Disruption: After font fallback, the original pattern of cluster value growth is broken, causing subsequent processing errors

## Solution
Considering the complexity of the issue and its impact on system stability, we chose to solve this problem at a higher level (in the shape_full_recursive function) rather than modifying the underlying makepad_rustybuzz implementation. The specific solution involves:

1. Adding a processing record mechanism to track processed text ranges
2. Checking if a position has been processed before handling each glyph
3. Avoiding repeated processing or processing invalid text ranges

This solution offers several advantages:
- Maintains compatibility with the HarfBuzz algorithm
- Doesn't increase overall system complexity
- Effectively resolves the processing order confusion
- Provides a reliable error handling mechanism

@ejpbruel2 

**Following the solution approach above, my fix code has resolved this issue in local testing. The code needs further organization before submission. I will share my analysis and solution approach first for discussion to see if it is the correct approach.**

ZhangHanDong:
@ejpbruel2 

The fix code has been submitted, please review.

Attached is a debugging log that I printed. Here is the processing procedure printed by the shape_full_ function when “赛！？” was entered.

The key to solving the problem is the introduction of a new parameter: base_cluster, which is used to pass the current cluster index value after the font switch, ensuring the continuity of the index value increment. Thereby solving the problem.

```
=== Begin Text Processing ===
Input text: "赛！？"
Text byte length: 9
Character count: 3
Available font count: 2
Base cluster offset: 0
Character Boundary Analysis:
Character[0]: '赛' Position=0 (Byte length=3)
Unicode: U+8D5B
Character type: Letter
Character[1]: '！' Position=3 (Byte length=3)
Unicode: U+FF01
Character type: Other
Character[2]: '？' Position=6 (Byte length=3)
Unicode: U+FF1F
Character type: Other
Obtained Glyph Information:
Total glyphs: 3
Glyph[0]:
cluster: 0
glyph_id: 0
status: Missing
absolute cluster position: 0
Glyph[1]:
cluster: 3
glyph_id: 0
status: Missing
absolute cluster position: 3
Glyph[2]:
cluster: 6
glyph_id: 0
status: Missing
absolute cluster position: 6
--- Processing Glyph ---
Relative cluster: 0
Absolute cluster: 0
glyph_id: 0
Processing missing glyph:
Found next valid position: 9
Attempting font fallback:
Slice range: [0..9]
Slice content: "赛！？"
Remaining fonts: 1
Transmitted base_cluster: 0
=== Begin Text Processing ===
Input text: "赛！？"
Text byte length: 9
Character count: 3
Available font count: 1
Base cluster offset: 0
Character Boundary Analysis:
Character[0]: '赛' Position=0 (Byte length=3)
Unicode: U+8D5B
Character type: Letter
Character[1]: '！' Position=3 (Byte length=3)
Unicode: U+FF01
Character type: Other
Character[2]: '？' Position=6 (Byte length=3)
Unicode: U+FF1F
Character type: Other
Obtained Glyph Information:
Total glyphs: 3
Glyph[0]:
cluster: 0
glyph_id: 26639
status: Valid
absolute cluster position: 0
Glyph[1]:
cluster: 3
glyph_id: 34042
status: Valid
absolute cluster position: 3
Glyph[2]:
cluster: 6
glyph_id: 34072
status: Valid
absolute cluster position: 6
--- Processing Glyph ---
Relative cluster: 0
Absolute cluster: 0
glyph_id: 26639
Processing valid glyph:
Adding to result set: font_id=2, glyph_id=26639, cluster=0
--- Processing Glyph ---
Relative cluster: 3
Absolute cluster: 3
glyph_id: 34042
Processing valid glyph:
Adding to result set: font_id=2, glyph_id=34042, cluster=3
--- Processing Glyph ---
Relative cluster: 6
Absolute cluster: 6
glyph_id: 34072
Processing valid glyph:
Adding to result set: font_id=2, glyph_id=34072, cluster=6
=== Processing Complete ===
Result set size: 3
Final glyph list:
Result[0]: font_id=2, glyph_id=26639, cluster=0
Result[1]: font_id=2, glyph_id=34042, cluster=3
Result[2]: font_id=2, glyph_id=34072, cluster=6
Font fallback successful
=== Processing Complete ===
Result set size: 3
Final glyph list:
Result[0]: font_id=2, glyph_id=26639, cluster=0
Result[1]: font_id=2, glyph_id=34042, cluster=3
Result[2]: font_id=2, glyph_id=34072, cluster=6
```

draw/src/font_atlas.rs
```diff
@@ -286,56 +287,95 @@ impl CxShapeCache {
...
         buffer.push_str(text);
+
+        // Shape the text using HarfBuzz.
         let buffer = font.owned_font_face.with_ref(|face| {
             makepad_rustybuzz::shape(face, &[], buffer)
         });
 
         let infos = buffer.glyph_infos();
+
+        // Modify the way glyph iteration is handled.
```
ejpbruel2: Is this supposed

draw/src/font_atlas.rs
```diff
@@ -286,56 +287,95 @@ impl CxShapeCache {
...
+                let mut preview_iter = info_iter.clone();
+                let next_valid = preview_iter
+                    .find(|next| next.glyph_id != 0)
+                    .map(|next| next.cluster as usize)
+                    .unwrap_or(text.len());
+                next_valid
+            };
+
+            // Verify the validity of the slice range.
+            if start >= next_cluster {
```
ejpbruel2: Why is this check necessary? Under what circumstances could it fail?

draw/src/font_atlas.rs
ejpbruel2: This is not a very useful comment. It makes sense in the context of this review, but people who will look a this code later will wonder *how* the glyph iteration was modified and why.

ejpbruel2 (commented):
In general, this code looks like it solves the problem correctly. Well done!

I just have two minor questions here below. Please address those, and then this should be good to land.

draw/src/font_atlas.rs
ZhangHanDong: This comment should be fine, right?
```
// Track the processed text position to avoid reprocessing characters
// that have already been handled through font fallback
let mut skip_to = 0;
```

draw/src/font_atlas.rs
ZhangHanDong: This is a defensive check.

Whether this check is still necessary depends on whether we can guarantee that `start` is always less than `next_cluster`.

The generation of cluster values is controlled by the `HarfBuzz` algorithm (`makepad_rustybuzz`). In theory, `HarfBuzz` should guarantee that cluster values are monotonically increasing. 

However, I am not sure if it is always monotonically increasing here, so I made a check as a robust implementation. 

This check is a simple comparison operation with very little overhead, but if we can be sure that this situation will never occur, this check is redundant and can be removed.

ZhangHanDong: Or convert this check into a debug assertion, so that potential issues can still be captured during the development phase.
```rust
debug_assert!(
    start < next_cluster,
    "HarfBuzz guarantees monotonic cluster values"
);
```

ZhangHanDong: > Or convert this check into a debug assertion, so that potential issues can still be captured during the development phase.或者将此检查转换为调试断言，以便在开发阶段仍然可以捕获潜在的问题。
> 
> ```rust
> debug_assert!(
>     start < next_cluster,
>     "HarfBuzz guarantees monotonic cluster values"
> );
> ```

 I changed it to this debug assertion, and it has been committed.

draw/src/font_atlas.rs
ZhangHanDong: > This comment should be fine, right?这个评论应该没问题吧？
> 
> ```
> // Track the processed text position to avoid reprocessing characters
> // that have already been handled through font fallback
> let mut skip_to = 0;
> ```

 I replaced it with this comment, and it has been committed.

PR #613, opened 2024-12-13
Add  `Ctrl + Z `shortcut to the input.

Guocork:
After some investigation, I found that on many platforms, the shortcuts for undo and redo are `Ctrl + Z` and `Ctrl + Shift + Z`. So far, I haven’t seen any platform (I tested on Windows, Linux, and Web) that supports` Logo + Z` or `Logo + Shift + Z`. Should we still keep the `Logo + Z` shortcut? Also, I noticed that on some apps or platforms (especially on the Web), `Ctrl + Y` is used for redo. Should Makepad also support the `Ctrl + Y` shortcut?

### DIFF
--- a/widgets/src/text_input.rs
+++ b/widgets/src/text_input.rs
@@ -640,7 +640,15 @@ impl Widget for TextInput {
                     ..
                 },
                 ..
-            }) if !self.is_read_only => {
+            }) | Hit::KeyDown(KeyEvent { 
+                key_code: KeyCode::KeyZ, 
+                modifiers: KeyModifiers {
+                    control: true,
+                    shift: false,
+                    ..
+                },
+                ..
+            }) if !self.is_read_only => { 
                 self.undo();
                 self.draw_bg.redraw(cx);
                 cx.widget_action(uid, &scope.path, TextInputAction::Change(self.text.clone()));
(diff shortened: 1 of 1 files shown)


kevinaboos:
for this, you can use my new addition of the `is_primary()` function to the `KeyModifier` struct, which will correctly return whether `Ctrl` is pressed on Windows/Linux or `Logo` is pressed on macOS.

See the PR here: https://github.com/makepad/makepad/pull/590

kevinaboos:
https://github.com/makepad/makepad/pull/590

Guocork:
> for this, you can use my new addition of the `is_primary()` function to the `KeyModifier` struct, which will correctly return whether `Ctrl` is pressed on Windows/Linux or `Logo` is pressed on macOS.
> 
> See the PR here: #590

Ok, thank you.I looked into it and found that the undo shortcut is `Command+Z`/ `Logo + Z `on macOS(cmd = logo in Mac), while `Ctrl+Z` is more common on non-macOS platforms, which better aligns with user habits.

kevinaboos (changes requested):
Currently, this code activates an Undo operation if either <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Z</kbd>  or <kbd>Logo</kbd> + <kbd>Shift</kbd> + <kbd>Z</kbd> is pressed. IMO, this is incorrect as the former is unexpected behavior on macOS, while the latter is unexpected on Linux/Windows.

Instead, this should activate an Undo operation only if  `KeyModifiers::is_primary()` is true and <kbd>Shift</kbd> + <kbd>Z</kbd> are pressed.

Guocork:
> Currently, this code activates an Undo operation if either Ctrl + Shift + Z or Logo + Shift + Z is pressed. IMO, this is incorrect as the former is unexpected behavior on macOS, while the latter is unexpected on Linux/Windows.
> 
> Instead, this should activate an Undo operation only if `KeyModifiers::is_primary()` is true and Shift + Z are pressed.

Ok, I got it and fix it, Thank you.

kevinaboos (approved):
LGTM

PR #614, opened 2024-12-17
Fix `Timer` error on linux

Guocork:
According to the timer `VecDeque` update method, the timer that's always released is the first one in the `VecDeque`. When the timer at `index=0` is released, its subsequent timers shouldn't have their `delta_timeout` increased. The error seems to be caused by this, because the trigger time of the timer is much longer than the set time. During the execution of the `stop_timer` method, unnecessary time was added to the `delta_timeout `of the second timer, causing the delay to be too long. I set up a timer in the simple module for testing purposes.I tested it on my computer, and it works fine.

### DIFF
--- a/examples/simple/src/app.rs
+++ b/examples/simple/src/app.rs
@@ -1,4 +1,6 @@
 
+use std::time::Instant;
+
 use makepad_widgets::*;
 
 live_design!{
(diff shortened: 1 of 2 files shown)


examples/simple/src/app.rs
yulnr, line 1: Hi @Guocork thanks for working on this!
The changes look good but I'm yet to test them. 
I think you might have committed this file (`app.rs`) by mistake, I understand you might've used this for testing, but we don't want to modify the simple example, could you revert this? :)

Guocork: Ok, I've reverted the changes to app.rs as requested. Please let me know if there's anything else needed. 😊

yulnr: Thanks, I just tested and it seems correct, good job!

yulnr (approved):

PR #621, opened 2024-12-26
CommandTextInput widget

noxware:
# Description

A widget that displays a popup with a list of options when certain key is pressed in its wrapped text input.

Options may be something mentionable (like people or bots, typically by detecting `@`), or maybe actual commands to execute (typically by detecting `/`).

# Notes

This can't be considered an autocomplete widget because when the popup appears, focus is interrupted. This mimics ChatGPT `@` behavior. That's why it's just called `CommandTextInput` and not `AutocompleteTextInput`.

This is a reusable version of [Moly's prompt input](https://github.com/moxin-org/moly/blob/a77b7f0221f95eccb668fdc4ce3bb76a79363f5e/src/chat/prompt_input.rs).

Commit history has been squashed due to a lot of experimentation and `wip` commits on it. 

# Simple example

```rust
use makepad_widgets::*;

live_design!(
    use link::theme::*;
    use link::widgets::*;

    pub Ui = {{Ui}} {
        item_template: <View> {
            height: 40,
            label = <Label> {}
        }

        body = <View> {
            flow: Down,
            align: {x: 0.5, y: 0.5}
            prompt = <CommandTextInput> { trigger: "/" }
        }
    }
);

#[derive(Live, LiveHook, Widget)]
pub struct Ui {
    #[deref]
    deref: Window,

    #[live]
    item_template: Option<LivePtr>,
}

impl Widget for Ui {
    fn draw_walk(&mut self, cx: &mut Cx2d, scope: &mut Scope, walk: Walk) -> DrawStep {
        self.deref.draw_walk(cx, scope, walk)
    }

    fn handle_event(&mut self, cx: &mut Cx, event: &Event, scope: &mut Scope) {
        self.deref.handle_event(cx, event, scope);

        let Event::Actions(actions) = event else {
            return;
        };

        let mut prompt = self.command_text_input(id!(prompt));

        if let Some(selected) = prompt.item_selected(actions) {
            let text = selected.label(id!(label)).text();
            println!("Selected: {}", text);
        }

        if prompt.should_build_items(actions) {
            prompt.clear_items();

            let search = prompt.search_text().to_lowercase();

            for fruit in ["Apple", "Banana", "Orange"] {
                if fruit.to_lowercase().contains(&search) {
                    let widget = WidgetRef::new_from_ptr(cx, self.item_template);
                    widget.label(id!(label)).set_text(fruit);
                    prompt.add_item(widget);
                }
            }
        }
    }
}
```

For more complex scenarios:
- Custom headings and grouping logic can be implemented by using `add_unselectable_item`.
- Since you are in control of the filtering, you can do something smarter than just comparing the lowercased strings.
- Usage in Moly can serve as a more [complex example](https://github.com/moxin-org/moly/pull/332).

### DIFF
--- a/widgets/src/command_text_input.rs
+++ b/widgets/src/command_text_input.rs
@@ -15,8 +15,8 @@ live_design! {
         flow: Down,
         height: Fit,
 
-        keyboard_focus_color: #ffffff3a,
-        pointer_hover_color: #ffffff23,
+        keyboard_focus_color: (THEME_COLOR_CTRL_HOVER),
+        pointer_hover_color: (THEME_COLOR_CTRL_HOVER * 0.85),
 
         popup = <RoundedView> {
             flow: Down,
--- a/widgets/src/command_text_input.rs
+++ b/widgets/src/command_text_input.rs
@@ -1,0 +1,610 @@
+use crate::*;
+
+live_design! {
+    link widgets;
+    use link::widgets::*;
+    use link::theme::*;
+
+    List = {{List}} {
+        flow: Down,
+        width: Fill,
+        height: Fill,
+    }
+
+    pub CommandTextInput = {{CommandTextInput}} {
+        flow: Down,
+        height: Fit,
+
+        keyboard_focus_color: #ffffff3a,
+        pointer_hover_color: #ffffff23,
+
+        popup = <RoundedView> {
+            flow: Down,
+            height: Fit,
+            visible: false,
+            search_input = <TextInput> {
+                width: Fill,
+                height: Fit,
+            }
+
+            list = <List> {
+                height: Fit
+            }
+        }
+
+        persistent = <RoundedView> {
+            flow: Down,
+            height: Fit,
+            top = <View> { height: Fit }
+            center = <RoundedView> {
+                height: Fit,
+                // `left` and `right` seems to not work with `height: Fill`.
+                left = <View> { width: Fit, height: Fit }
+                text_input = <TextInput> { width: Fill }
+                right = <View> { width: Fit, height: Fit }
+            }
+            bottom = <View> { height: Fit }
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


widgets/src/command_text_input.rs
```diff
@@ -0,0 +1,610 @@
...
+        width: Fill,
+        height: Fill,
+    }
+
+    pub CommandTextInput = {{CommandTextInput}} {
+        flow: Down,
+        height: Fit,
+
+        keyboard_focus_color: (THEME_COLOR_CTRL_HOVER),
+        pointer_hover_color: (THEME_COLOR_CTRL_HOVER * 0.85),
```
noxware, line 19: Does it make sense to use this color? Or is there a better one?

widgets/src/command_text_input.rs
```diff
@@ -0,0 +1,610 @@
...
+                width: Fill,
+                height: Fit,
+            }
+
+            list = <List> {
+                height: Fit
+            }
+        }
+
+        persistent = <RoundedView> {
```
noxware, line 35: Adding a couple of empty View slots here to simplify placement of stuff around the main text input. Without this, Moly's layout would be complex to recreate.

widgets/src/command_text_input.rs
```diff
@@ -0,0 +1,610 @@
...
+live_design! {
+    link widgets;
+    use link::widgets::*;
+    use link::theme::*;
+
+    List = {{List}} {
+        flow: Down,
+        width: Fill,
+        height: Fill,
+    }
```
noxware, line 12: Brought a simplified copy of the `List` widget from Moly, as `PromptInput` (the source of this widget) was implemented on top of it.

It's possible to use `FlatList` instead, but it may leak memory. And `PortalList` hasn't been adapted to support `FlatList` use cases yet.

PR #632, opened 2025-01-08
CommandTextInput chinese fixes

noxware:
# Changes
- Fix trigger by Chinese grapheme.
  - By working with graphemes instead of chars to handle triggers composed of multiple unicode code points.
- Fix trigger detection after text of the input has been changed programmatically.
  - By using `Event::TextInput` to get the inserted text instead of tracking the previous text to do so.
- Some other minor improvements and some simplifications.


widgets/src/command_text_input.rs
```diff
@@ -267,28 +255,36 @@ impl CommandTextInput {
...
-                self.previous = at_removed;
-            }
+        let text = self.text();
+        let Some((inserted_grapheme_pos, inserted_grapheme)) =
+            inserted_grapheme_with_pos(&text, head)
+        else {
+            return;
+        };
+
+        if Some(inserted_grapheme) == self.trigger_grapheme() {
```
kevinaboos: nit: it's more readable and Rust-y to flip it, otherwise it looks like a mistaken `if let Some(...)`.
```suggestion
        if self.trigger_grapheme() == Some(inserted_grapheme) {
```

widgets/src/command_text_input.rs
```diff
@@ -239,8 +230,8 @@ impl Widget for CommandTextInput {
 }
 
 impl CommandTextInput {
-    fn on_text_input_changed(&mut self, cx: &mut Cx, scope: &mut Scope) {
-        if self.was_trigger_inserted() {
+    fn on_text_input_insert(&mut self, cx: &mut Cx, scope: &mut Scope, inserted: &str) {
```
kevinaboos: nit: a clearer name might be `on_text_inserted` or `on_insert_into_text_input`

kevinaboos (changes requested):
just a few minor nitpicks

kevinaboos (approved):

PR #636, opened 2025-01-16
`HitOptions`: add a callback to determine whether a `Hit` should be set as "handled"

kevinaboos:
If the new option is `None` (not set), the existing behavior is preserved, which is to mark the `Hit` as `handled`.
Marking a `Hit` as "handled" means that it will not be delivered to child widgets.

If the callback returns `false`, the `Hit` is not marked as handled. This is useful when you want to allow a `Hit` to be handled in a given widget AND to be handled by that widgets' child widgets.


kevinaboos:
Closed in favor of the new `Event::unhandle()`function.

PR #637, opened 2025-01-17
Introduce new `MouseButton` bitmask/bitflags to represent pressed mouse buttons

kevinaboos:
**All good and ready to merge!**

~~Not yet ready to merge~~

---------------------

The `MouseButton` type is intended to provide a multi-platform abstraction for the 5 common mouse buttons, as well as support for arbitrary button numbers. These 5 buttons are:
1. The primary mouse button, typically the left-click button.
2. The secondary mouse button, typically the right-click button.
3. The middle mouse button, typically the scroll-wheel click button.
4. The fourth mouse button, typically used for back navigation.
5. The fifth mouse button, typically used for forward navigation.

It also supports the notion of multiple mouse buttons being pressed simultaneously, e.g., left-clicking while the right-click button is held down (though many platforms don't actually support emitting those events).

Platforms currently supported:
- [x] macOS
- [x] windows 
- [x] Linux X11
-  ~~Android~~ no specific mouse support
- [x] iOS
- [x] Linux direct
- [x] web
- ~~ohos~~ (build is broken)

### DIFF
--- a/platform/src/event/finger.rs
+++ b/platform/src/event/finger.rs
@@ -59,14 +59,78 @@ impl KeyModifiers{
     }
 }
 
-#[derive(Clone, Debug, PartialEq)]
+
+bitflags::bitflags! {
+    /// A bit mask of all mouse buttons that were pressed
+    /// during a given mouse event.
+    ///
+    /// This is a bit mask because it is possible for multiple buttons
+    /// to be pressed simultaneously during a given input event.
+    #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+    pub struct MouseButton: u32 {
+        /// The primary mouse button, typically the left-click button.
+        const PRIMARY =   1 << 0;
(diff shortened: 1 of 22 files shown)


platform/src/event/finger.rs
```diff
@@ -1064,7 +1143,7 @@ impl Event {
                 if cx.fingers.test_sweep_lock(options.sweep_area) {
                     return Hit::Nothing;
                 }
-                let device = DigitDevice::Mouse { button: 0 };
+                let device = DigitDevice::Mouse { button: MouseButton::empty() };
```
kevinaboos, line 1146: @makepaddev is this correct? Previously, a button value of `0` meant the primary button, but that seems weird to me. Thus, I used a "no" button bitmask instead, which indicates that no button was pressed.

I'm not quite sure why a `MouseLeave` event needs a button press, but do let me know if we need to switch it to `MouseButton::PRIMARY`.

PR #651, opened 2025-02-01
Modify the visibility of CommandTextInput field

ZhangHanDong:
Set the visibility of the `keyboard_focus_index` field.

```rust
pub keyboard_focus_index: Option<usize>,
```

### DIFF
--- a/widgets/src/command_text_input.rs
+++ b/widgets/src/command_text_input.rs
@@ -106,7 +106,7 @@ pub struct CommandTextInput {
     /// Index from `selectable_widgets` that would be submitted if `Return` is pressed.
     /// `None` if there are no selectable widgets.
     #[rust]
-    keyboard_focus_index: Option<usize>,
+    pub keyboard_focus_index: Option<usize>,
 
     /// Index from `selectable_widgets` that the pointer is hovering over.
     /// `None` if there are no selectable widgets.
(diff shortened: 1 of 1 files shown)


widgets/src/command_text_input.rs
```diff
@@ -106,7 +106,7 @@ pub struct CommandTextInput {
     /// Index from `selectable_widgets` that would be submitted if `Return` is pressed.
     /// `None` if there are no selectable widgets.
     #[rust]
-    keyboard_focus_index: Option<usize>,
+    pub keyboard_focus_index: Option<usize>,
```
noxware: Ah, this field visibility you meant. Makes sense.

If possible, I would avoid making this field public as it's internal and setting it to a wrong value, like `None` while there are options, `Some(usize)` when there aren't any options, or an out of bounds index, can cause undefined behavior. 

I propose the following methods (or similar) as an alternative:

```rust
pub fn keyboard_focus_index(&self) -> Option<usize> {
    self.keyboard_focus_index
}

pub fn set_keyboard_focus_index(&mut self, idx: usize) {
    if self.selectable_widgets.len() > 0 {
        self.keyboard_focus_index = Some(idx.clamp(0, self.selectable_widgets.len() - 1));
    };
}
```

(I can't test it right now 😅 but that should be the idea)

ZhangHanDong: Agree.

noxware (approved):

PR #654, opened 2025-02-08
Rik cleanup

NinekoTheCat:
This pull request:
- Cleans up code in `image_cache.rs` to make it more readable
- Adds `.gitattributes` to normalize file line endings properly
- fixes bug where previous assertion about `from_png` changed causing any images that were not pngs passed in to panic.
    - Previous assertion was that passed in images that are not png will result in an Err being returned

### DIFF
--- a/libs/zune-png/src/lib.rs
+++ b/libs/zune-png/src/lib.rs
@@ -162,6 +162,7 @@ extern crate alloc;
 extern crate core;
 
 #[cfg(feature = "std")]
+pub use apng::ActlChunk;
 pub use apng::post_process_image;
 pub use apng::{BlendOp, DisposeOp};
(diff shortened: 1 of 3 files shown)


libs/zune-png/src/lib.rs
```diff
@@ -162,6 +162,7 @@ extern crate alloc;
 extern crate core;
 
 #[cfg(feature = "std")]
+pub use apng::ActlChunk;
```
univerz: adding a line before one protected by conditional compilation looks suspicious

NinekoTheCat: Oh my god, I am so sorry. the rust language server did that

NinekoTheCat: https://github.com/makepad/makepad/pull/654/commits/e4d501d114b9d5d8dce00a760adbef037106eb4a fixes this

PR #673, opened 2025-03-08
Breaking change: allow app logic to mark a `Hit` event as [un]handled

kevinaboos:
(This PR depends on / includes changes from #670, so merge that one in first before trying to review this)

This commit adds a new paradigm that allows the app dev to control whether the act of handling hits from a given event (when calling `event.hits()`) will or will not mark that event as handled.

The default is that the `Hit` *does* mark the causing event as handled, such that it won't be handle-able by other widgets duplicately unless of course they enable `capture_overload`).
This commit still preserves that historical behavior in order to not surprise any users further; however, I do believe that the default should be that an event is *not* considered to be handled unless the app logic explicitly marks it as handled.

This currently only applies to four `Hit` variants:
* `Hit::FingerDown`
* `Hit::FingerMove`
* `Hit::FingerHoverIn`
* `Hit::FingerHoverOver`

because those are the four `Hit`s that previously set their event's `handled` Area unconditionally (internal to the `hits()` API).

Any code that matches on one of those four `Hit` variants will need to be updated. It's very simple to do so, you can just ignore the extra `EventHandled` parameter when matching on those variants.
```rust
pub enum Hit {
    ...
    FingerDown(FingerDownEvent, EventHandled),
    FingerMove(FingerMoveEvent, EventHandled),
    FingerHoverIn(FingerHoverEvent, EventHandled),
    FingerHoverOver(FingerHoverEvent, EventHandled),
    ...
```

In addition, all of these Hit handlers within the base `View` widget now explicitly mark those events as *not* handled, which ensures they are also able to handled by other widgets.
This makes sense because the base `View` should not "consume" any hits, since most app devs would not expect that a blank empty space view would consume a mouse click or something.

### DIFF
--- a/audio_graph/audio_widgets/src/piano.rs
+++ b/audio_graph/audio_widgets/src/piano.rs
@@ -209,14 +209,14 @@ impl PianoKey {
             self.draw_key.area().redraw(cx);
         }
         match event.hits_with_sweep_area(cx, self.draw_key.area(), sweep_area) {
-            Hit::FingerHoverIn(_) => {
+            Hit::FingerHoverIn(..) => {
                 cx.set_cursor(MouseCursor::Hand);
                 self.animator_play(cx, id!(hover.on));
             }
             Hit::FingerHoverOut(_) => {
                 self.animator_play(cx, id!(hover.off));
             }
-            Hit::FingerDown(_) => {
+            Hit::FingerDown(..) => {
                 self.animator_play(cx, id!(hover.on));
                 self.animator_play(cx, id!(pressed.on));
                 actions.push((key_id, PianoKeyAction::Pressed(127)));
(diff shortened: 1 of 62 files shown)


kevinaboos:
@joulei  would appreciate your review here since this will have a minor impact Moly and your other apps. (for some reason github won't let me request a review from you)

kevinaboos:
After some discussion, we'll take a different approach to re-implementing hit handling and capture undoing:

> * add unhandle to fingerdown
> * add an area 'vec' for capture_overloads
> * make hits fire again for the same Area if that is a handled area to allow deref view hit handling

kevinaboos:
Closed in favor of the new `Event::unhandle()`function.

PR #681, opened 2025-03-13
Add  `set_url` and `url()` public function to HtmlLink widget.

tyreseluo:


kevinaboos (approved):
Looks good, but FYI this is already available via my changes in #666

tyreseluo:
Thanks. Yes, although the HtmlLink structure is set to public, we still have to add the relevant API for HtmlLinkRef.

I have changed the code.

kevinaboos (approved):

PR #685, opened 2025-03-17
Resize and reposition window [Macos]

alanpoon:
Add the ability to resize and reposition window.

 if self.ui.button(id!(button1)).clicked(&actions) {
             self.counter += 1;
             // This does not work
             //self.ui.window(id!(main_window)).resize(cx, DVec2{x: 1600.0, y: 800.0});
            // Resize and reposition work
             cx.push_unique_platform_op(CxOsOp::ResizeWindow(WindowId(0,0), DVec2{x: 1600.0, y: 800.0}));
             cx.push_unique_platform_op(CxOsOp::RepositionWindow(WindowId(0,0), DVec2{x: -400.0, y: 0.0}));
}


kevinaboos:
Nice! This seems to only support macOS for now, so you should add that to the PR title & description too.

PR #711, opened 2025-04-10
fix overriding dock keys after loading state

alanpoon:
https://github.com/project-robius/robrix/pull/422#issuecomment-2787981245

A typical use for dock load and dock clone, is to restore dock panel after the app is restarted.
In a new instance of Makepad, dock drag and move overwrites the old splitter after dock load_state because 
it is calling LiveId::unique() which is back to 1 in the new split.

Hence, causing old split to be overwritten by the new split in the hash map.

As can be seen in this video:
https://github.com/user-attachments/assets/4946a500-d90a-43fc-82a9-f5552dda4ad1
Here is my test branch to reproduce the issue https://github.com/alanpoon/makepad/tree/dock_load_state_fix_test. 
`UNIQUE_LIVE_ID.store(1, Ordering::SeqCst);` is added before dock.load_state to simulate new instance of makepad.

The solution is to change all unique id into other strings before loading into the dock state.


kevinaboos:
nice work!

widgets/src/dock.rs
```diff
@@ -907,8 +907,39 @@ impl Dock {
...
+                                *t = LiveId::from_str(&format!("Old Unique {:?}", t.0));
+                            }
+                        })
+                    }
+                    _ => {}
+                };
+                (new_key, v)
+            })
+            .collect();
         //log!("{:#?}", self.dock_items);
```
ZhangHanDong, line 941: After reviewing the code, it's suggested to define the prefix as a constant: `const OLD_UNIQUE_ID_PREFIX: &str = "Old Unique ";  `  This slightly improves maintainability.

ZhangHanDong (commented):
Additionally, in the current solution, the original numeric unique LiveId is discarded. While this doesn't seem to be a problem, is there any special case (which I can't think of at the moment) where discarding the original LiveId could lead to errors?

widgets/src/dock.rs
makepaddev: LiveIds do unique() to create new 'runtime' unique ids. It's simply a global atomic counter.
If you save/load them from disk you need keep track of the highest UniqueLiveID in the loaded set,
And then make sure UNIQUE_LIVE_ID global is set to that +1, then you will never have collisions.
This is libs/live_id/src/live_id.rs line 278
    pub fn unique() -> Self {
        LiveId(UNIQUE_LIVE_ID.fetch_add(1, Ordering::SeqCst))
    }

makepaddev: Of course this does assume that you dont 'keep' tabs around from the dock before. Although this solution you have here also works.

makepaddev: This solution seems fine

makepaddev:
I have removed it again after i looked into how i do things in studio.
You call DockRef::unique_tab_id on line 1241 of dock.rs to create new tab ids, dont use LiveId::unique!

alanpoon:
> I have removed it again after i looked into how i do things in studio. You call DockRef::unique_tab_id on line 1241 of dock.rs to create new tab ids, dont use LiveId::unique!

This bug can also been replicated in studio.
https://github.com/user-attachments/assets/144f2aea-0684-4f86-8afa-e7cfa9a73a6e

makepaddev:
Thats weird, unique tab should always return a unique tab id for the current state.

alanpoon:
https://github.com/makepad/makepad/blob/6335300aafc43551b6046bf57fbbdc9a3e5be45d/widgets/src/dock.rs#L628
LiveId::unique is used in inner function "handle_drop", not the public functions of the DockRef.

makepaddev:
Ahh ok thats a bug then although im not sure the tabs and splitters use the same ID space, but ifso then this is a bug.

kevinaboos:
@alanpoon can you submit a new PR that implements this in a way that shows (and fixes) the problem with Makepad studio as well? Once that solution is merged in, then we can use it for https://github.com/project-robius/robrix/pull/422.

PR #724, opened 2025-04-25
Restore dock collision fix

alanpoon:
#711 

https://github.com/user-attachments/assets/41f9504d-57cf-4254-bf3a-416116c5fc72

### DIFF
--- a/widgets/src/dock.rs
+++ b/widgets/src/dock.rs
@@ -7,7 +7,6 @@ use crate::{
     splitter::{SplitterAction, Splitter, SplitterAlign, SplitterAxis},
     tab_bar::{TabBarAction, TabBar},
 };
-const OLD_UNIQUE_ID_PREFIX: &str = "Old Unique ";
 
 live_design!{
     link widgets;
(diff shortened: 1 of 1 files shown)


kevinaboos (changes requested):
for the sake of efficiency, do we need the prefix to be so many bytes? And does it have to be a string?

Seems like it could just be a single byte added onto the existing non-unique LiveId, which you could realize with something like `existing_live_id.bytes_append(...)` or another function.
We shouldn't need to run a string format routine (which requires allocation) to generate a long string that just becomes hashed immediately after.

makepaddev:
I have fixed the uniqueness of the tab bar and splitter ids as well

kevinaboos:
> I have fixed the uniqueness of the tab bar and splitter ids as well

Thanks! can you link the commit here so we can see the changes?

kevinaboos:
looks like it was this: https://github.com/makepad/makepad/commit/5195934e5ce1e6896bceffdd0e0226e0d09af895

@alanpoon can you confirm that this works for us to fix this? https://github.com/project-robius/robrix/pull/422#issuecomment-2787981245

PR #726, opened 2025-04-29
Add more window function for windowRef

alanpoon:
1. get_position
2. is_fullscreen
3. fullscreen


makepaddev:
Btw i know we are not 100% consistent here but rustidom is:
name() == getter
set_name() == setter

PR #729, opened 2025-05-02
Add `Dock::replace_tab()`. Fix StackNavigation to forward non-visibility events to all subviews

kevinaboos:
The new `Dock::replace_tab()` function allows a user to replace the inner content of a tab without closing or recreating that tab. 

See docs copied below:
```rust
    /// Performs an in-place replacement of the inner widget (within an existing tab)
    /// with a completely new widget of a different kind.
    /// The type of the tab itself is not changed, but the widget inside of it is.
    ///
    /// Optionally, you can give the tab a new name and select it.
    ///
    /// ## Arguments
    /// * `cx`: A mutable reference to the Cx context.
    /// * `tab_item_id`: The ID of the tab to be replaced.
    /// * `new_kind`: The new widget that should be shown inside of the tab.
    /// * `new_name`: An optional new name for the tab.
    /// * `select`: A boolean indicating whether to select the new tab after replacement.
    ///   If `false`, the tab will be not be explicitly selected, but of course
    ///   if it was already selected, it will remain selected.
    ///
    /// ## Return
    /// * Returns `None` if the `tab_item_id` does not exist in the dock
    ///   or if the `new_kind` template was not listed in the Dock's live DSL.
    /// * Returns `Some((new_widget, true))` if the existing tab's inner widget
    ///   was successfully replaced with the new one. 
    ///   * Returns `Some((existing_widget, false))`` if the `new_kind` is the same
    ///     as existing tab's inner widget kind, meaning that no replacement occurred.
    fn replace_tab(
        &mut self,
        cx: &mut Cx,
        tab_item_id: LiveId,
        new_kind: LiveId,
        new_name: Option<String>,
        select: bool,
    ) -> Option<(WidgetRef, bool)> {
```

### DIFF
--- a/widgets/src/dock.rs
+++ b/widgets/src/dock.rs
@@ -95,7 +95,7 @@ live_design!{
     }
 
     pub DockToolbar = <RectShadowView> {
-        width: Fill, height: 38.,
+        width: Fill, height: 40.,
         flow: Down,
         align: { x: 0., y: 0. }
         margin: { top: -1. }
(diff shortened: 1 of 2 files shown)


kevinaboos:
Closing in favor of #734 and #735, since we're now targeting the `dev` branch.

PR #731, opened 2025-05-07
Allow saving and restoring the state of a `TextInput` widget

kevinaboos:
This redoes the changes introduced in #533, but modified for the new version of `TextInput`

I'm not sure if we absolutely need to include `laidout_text` and `text_area`, since those are new additions to the widget.


kevinaboos:
Closed in favor of #736

PR #736, opened 2025-05-08
Allow saving/restoring the state of the `TextInput` widget

kevinaboos:
This redoes the changes introduced in #533, but modified for the new version of TextInput


kevinaboos:
~~Actually this doesn't fully work correctly with the latest `TextInput` widget, so i'm marking it as a draft until i can fix it.~~

kevinaboos:
ok fixed now, ready to go.

PR #754, opened 2025-05-21
feat: Implement AudioPlayer widget for Android

ZhangHanDong:
!!! **This PR was submitted by AI and is only for verifying the implementation approach—it is not intended for final merging** !!!

This commit introduces a new AudioPlayer widget to makepad-widgets, enabling audio playback functionality, primarily targeting the Android platform using MediaPlayer.

Key features and changes:

1.  **AudioPlayer Widget (Rust)**:
    *   Provides a user-friendly API with properties like `source`, `auto_play`, `loop_playback`, `initial_volume`.
    *   Supports actions such as `Play`, `Pause`, `Stop`, `SeekTo`, `SetVolume`, `SetMute`.
    *   Emits events like `Prepared`, `PlaybackStarted`, `PlaybackCompleted`, `Error`.
    *   Includes a basic UI with a play/pause button, time display, and progress bar.
    *   Integrates with `live_design!` for configuration and dynamic updates.

2.  **Platform Layer (Rust)**:
    *   Defined `AudioSource` enum for specifying audio sources (URL, File, LiveDependency).
    *   Extended `CxOsOp` with new operations for audio playback control.
    *   Setup for handling JNI callbacks from the Android platform.

3.  **Android Implementation (Java)**:
    *   Created `AudioPlayer.java` wrapping `MediaPlayer` for playback logic.
    *   Utilizes a dedicated `HandlerThread` for audio operations.
    *   Implemented JNI bridges in `MakepadActivity.java` and `MakepadNative.java` for Rust-Java communication.

4.  **Robrix Example App**:
    *   Added a new example application (`examples/robrix`) to demonstrate `AudioPlayer` usage with a network audio source.

5.  **Documentation**:
    *   Included comprehensive Markdown documentation for the `AudioPlayer` widget in `makepad-widgets/src/audio_player/README.md`.

This implementation allows Makepad applications to easily incorporate audio playback. Future work could include extending support to macOS/iOS and adding features like `TimeUpdate` events from the platform.

### DIFF
--- a/widgets/src/audio_player.rs
+++ b/widgets/src/audio_player.rs
@@ -1,0 +1,432 @@
+use makepad_platform::{
+    makepad_live_id::{LiveId, live_id},
+    event::Event,
+    cx_api::CxOsOp,
+    audio::AudioSource,
+    live_hook::{LiveHook, LiveApply, ApplyFrom, LiveNode},
+    derive_live::Live,
+    cx::Cx,
+    area::Area,
+    layout::{Walk, Layout, Align, Padding, Flow},
+    draw_2d::{Cx2d, DrawShape, DrawText, Text},
+    widget::{Widget, WidgetDraw, WidgetActionItem, WidgetRef, WidgetSet, WidgetActions, ButtonAction, ButtonWidgetRefExt},
+    // scope_form_action_cast, // Not standard, remove if not defined elsewhere
+    scope::Scope,
+    Signal,
+    event::{
+        AudioPlaybackPreparedEvent,
+        AudioPlaybackStartedEvent,
+        AudioPlaybackPausedEvent,
+        AudioPlaybackStoppedEvent,
+        AudioPlaybackCompletedEvent,
+        AudioPlaybackErrorEvent,
+        AudioPlaybackReleasedEvent,
+        // AudioPlaybackTimeUpdateEvent, // Assuming this will be added to platform/src/event/event.rs if used
+        finger::{FingerDownEvent, FingerUpEvent, FingerHoverEvent, HoverState},
+    },
+    vec2,
+    Color,
+};
+use crate::button::Button; // Import the Button widget
+
+#[derive(Live, LiveHook, LiveApply, Widget)]
+pub struct AudioPlayer {
+    #[live] walk: Walk,
+    #[live] layout: Layout,
+    #[live] area: Area,
+
+    #[live] player_id: LiveId,
+
+    #[live(AudioSource::default())] source: AudioSource,
+    #[live(false)] auto_play: bool,
+    #[live(false)] loop_playback: bool,
+    #[live(1.0)] initial_volume: f64,
+    #[live(false)] initial_mute_state: bool,
+
+    #[rust] is_prepared: bool,
+    #[rust] is_playing: bool,
+    #[rust] duration_ms: Option<u64>,
+    #[rust] current_time_ms: Option<u64>,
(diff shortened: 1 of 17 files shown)


widgets/src/audio_player.rs
```diff
@@ -0,0 +1,432 @@
...
+                        cx.platform_ops.push(CxOsOp::SeekAudioPlayback(self.player_id, time_ms as f64));
+                        self.current_time_ms = Some(time_ms);
+                        self.area.redraw(cx);
+                    }
+                }
+                AudioPlayerAction::SetVolume(volume) => {
+                    let clamped_volume = volume.max(0.0).min(1.0);
+                    self.current_volume = clamped_volume;
+                    cx.platform_ops.push(CxOsOp::SetAudioVolume(self.player_id, clamped_volume));
+                }
```
yulnr, line 318: Just FYI, volume manipulation via volume keys (on the phone) is already handled automatically by Android, we have this configured in MakepadActivity with the line `setVolumeControlStream(AudioManager.STREAM_MUSIC);`.

I do understand you might want to handle other keys or programmatically manipulate volume and that is also ok.

widgets/src/audio_player.rs
```diff
@@ -0,0 +1,432 @@
...
+                    }
+                }
+                AudioPlayerAction::Resume => {
+                    if self.is_prepared { // Resume implies it was prepared and paused
+                         cx.platform_ops.push(CxOsOp::ResumeAudioPlayback(self.player_id));
+                    }
+                }
+                AudioPlayerAction::Stop => {
+                    cx.platform_ops.push(CxOsOp::StopAudioPlayback(self.player_id));
+                }
```
yulnr, line 306: In general I would avoid using actions to perform operations on the widget, and I would rather use methods in the WidgetRef, in this case `AudioPlayerRef`, where you would do things like `audio_player.play(...)`. 

I think in Makepad we generally favor using the widget actions as "widgets events" instead, for example what's done here for `PlaybackStarted`, `PlaybackPaused` etc (this is correct). 
One note here is that I would avoid naming this actions "event" to prevent confusing people (events and actions are very already confusing), so I would name `AudioPlayerEvent` -> `AudioPlayerAction` instead, this way we avoid confusing these actions with the platform events, e.g. `Event::AudioPlaybackPrepared`.

yulnr (commented):
Hi @ZhangHanDong thanks for working on this, the general approach here makes sense and it's close to what I had in mind.

I'm leaving some comments about the implementation. 

I've previously tested automatic audio routing (e.g. using the earpiece vs the general speaker), using Android's `AudioManager` API. You can do something like:
```java
  if (useEarpiece && hasEarpiece()) {
      mAudioManager.setMode(AudioManager.MODE_IN_COMMUNICATION);
      mAudioManager.setSpeakerphoneOn(false);
      Log.i("Makepad", "Audio output switched to earpiece");
  } else {
      mAudioManager.setMode(AudioManager.MODE_NORMAL);
      mAudioManager.setSpeakerphoneOn(true);
      Log.i("Makepad", "Audio output switched to speaker");
  }
```
And then some other utilities for checking the proximity sensors and automatically updating the audio routing, I have this in a draft and can share it later. Audio routing does not need to be done in the first iteration of the audio player, but I wanted to let you know that it's seems to be easily doable.

**Note about platform support**
I think we should include some warnings/logs in the widget whenever it is being used outside of android. That way we can warn devs that the widget is not expected to be usable outside Android, you can find examples of this in the Video widget.

PR #766, opened 2025-06-04
Rework StackNavigation into an actual Stack

yulnr:
Refactors the StackNavigation system into a proper **stack** with a more idiomatic API that matches patterns from other UI frameworks.

#### New API
- `push(cx, view_id)` - Navigate to a new view (replaces show_stack_view_by_id)
- `pop(cx)` - Go back to previous view
- `pop_to_root(cx)` - Return to root view

#### Proper Navigation Stack
- Maintains complete navigation history instead of tracking only one active view
- Fixes animation issue where transitions always appeared to come from root
- Prevents navigation cycles by removing duplicate views from stack

#### Some inspection methods
- `depth()` - Get current stack depth
- `can_pop()` - Check if back navigation is possible
- `current_view()` - Get current view ID
- `stack_view_ids()` - Get complete navigation history

Annotated `show_stack_view_by_id` with `#[deprecated]` therefore the only breaking change is `StackNavigationAction::NavigateTo` being entirely replaced with `StackNavigationAction::Push(view_id)`.

#### Updated actions
```rust
// Old actions
StackNavigationAction::NavigateTo(view_id)

// New actions  
StackNavigationAction::Push(view_id)
StackNavigationAction::Pop
StackNavigationAction::PopToRoot
```

#### Support for multiple concurrent StackNavigation's, and non-full screen
- A makepad app can now have multiple instances of StackNavigation, including nested ones.
- The navigation views (`StackNavigationView`) can now be flagged with `full_screen: false` (`true` by default), to constraint the views to the StackNavigation boundaries, and disable the sliding animations. I've came across a few limitations of the Makepad animator that prevent me from supporting the animations for both full-screen and non-full screen within the same widget (unable to parametrize or update the animator target values programmatically).
Disabling full screen can be useful when using multiple StackNavigations within the same view (e.g. side-to-side in a desktop layout).

### DIFF
--- a/widgets/src/stack_navigation.rs
+++ b/widgets/src/stack_navigation.rs
@@ -127,7 +127,12 @@ live_design!{
 #[derive(Clone, DefaultNone, Eq, Hash, PartialEq, Debug)]
 pub enum StackNavigationAction {
     None,
-    NavigateTo(LiveId)
+    /// Push a new view onto the navigation stack
+    Push(LiveId),
+    /// Pop the current view from the navigation stack
+    Pop,
+    /// Pop all views and return to the root view
+    PopToRoot,
 }
 
 #[derive(Clone, Default, Eq, Hash, PartialEq, Debug)]
@@ -210,7 +215,11 @@ impl StackNavigationView {
                 || matches!(event, Event::Actions(actions) if self.button(id!(left_button)).clicked(&actions))
                 || matches!(event, Event::MouseUp(mouse) if mouse.button.is_back())
             {
-                self.hide_stack_view(cx);
+                cx.widget_action(
+                    self.widget_uid(),
+                    &HeapLiveIdPath::default(),
+                    StackNavigationAction::Pop,
+                );
             }
         }
     }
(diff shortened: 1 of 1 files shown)


alanpoon:
Would you support multiple Stack navigation Widget?
https://github.com/alanpoon/makepad/tree/stack-navigation-back-to-root-multiple/examples/stack_navigation_example/src

I have 2 stack navigation Widgets. Opening the left stacknavigation view, pushes the right stack navigation to the left. 

https://github.com/user-attachments/assets/9bcb3c26-53d6-4dfb-9bce-e485427db0f7

yulnr:
> Would you support multiple Stack navigation Widget? https://github.com/alanpoon/makepad/tree/stack-navigation-back-to-root-multiple/examples/stack_navigation_example/src
> 
> I have 2 stack navigation Widgets. Opening the left stacknavigation view, pushes the right stack navigation to the left.
> Screen.Recording.2025-06-20.at.11.39.17.AM.mov

Hi @alanpoon, yes this is a current limitation of stacknavigation due to how it uses actions to communicate between the navigation views and the main stack navigation widget. Currently there's no proper checking of the views belonging to the right stack navigation.
It shouldn't be hard to fix, I'll take a look tomorrow.

yulnr:
Hi @alanpoon I've updated the PR adding support multiple instances of StackNavigation, and also support for non-full screen (like in the video you attached) for that use case you'll have to flag your `StackNavigationView` with `full_screen: false` in the DSL, as the default behaviour is still to take over the screen. I left more details at the end of the PR description.

Here's a gist with [working example](https://gist.github.com/joulei/960634ef74eef34a3c4b1d579f62db5d) similar to the one you shared.

PR #768, opened 2025-06-08
fix image_size_by_path() to support PNG

Guocork:


Guocork:
Extend the `vec` to accommodate the size of the PNG image and use `read_to_end` to ensure all data is read completely.

kevinaboos:
I think the point of the original code is that you only read the first chunk of bytes (instead of the entire image) in order to decode the image header and determine its size.

Do PNGs require reading the entire file in order to decode it?

makepaddev:
Its fixed already

PR #770, opened 2025-06-13
Persistent and restoration of Window State

alanpoon:
- Allow creation of window using function instead of using live_design.
- Remove is_created check inside get_inner_size and get_position function

### DIFF
--- a/platform/src/window.rs
+++ b/platform/src/window.rs
@@ -181,16 +181,12 @@ impl WindowHandle {
         cx.windows[self.window_id()].main_pass_id = Some(pass.pass_id());
         cx.passes[pass.pass_id()].parent = CxPassParent::Window(self.window_id());
     }
-    pub fn create_window(&mut self, cx: &mut Cx, inner_size: DVec2, position: DVec2, is_fullscreen: bool, title: String) {
-        let window = &cx.windows[self.window_id()];
+    pub fn configure_window(&mut self, cx: &mut Cx, inner_size: DVec2, position: DVec2, is_fullscreen: bool, title: String) {
+        let window = &mut cx.windows[self.window_id()];
         window.create_title = title;
         window.create_position = Some(position);
         window.create_inner_size = Some(inner_size);
-        window.is_created = true;
-        cx.push_unique_platform_op(CxOsOp::CreateWindow(self.window_id()));
-        if is_fullscreen && self.can_fullscreen(cx){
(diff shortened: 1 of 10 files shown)


platform/src/window.rs
```diff
@@ -181,7 +181,16 @@ impl WindowHandle {
         cx.windows[self.window_id()].main_pass_id = Some(pass.pass_id());
         cx.passes[pass.pass_id()].parent = CxPassParent::Window(self.window_id());
     }
-    
+    pub fn create_window(&mut self, cx: &mut Cx, inner_size: DVec2, position: DVec2, is_fullscreen: bool, title: String) {
```
kevinaboos: naming question: is this actually _creating_ the window? Seems like it isn't, because it already expects the window to exist in `cx.windows`. I realize it pushes a `CxOsOp::CreateWindow`, so the name may be appropriate, but the precondition is a bit vague. Other naming options include words like "configure" or "initialize". 
 
Either way, I would recommend documenting the exact preconditions for calling this:
* when can it be called (after what point in the app/window initialization flow)?
* when should it _not_ be called? Someone might be tempted to use this function to set various window parameters, but because it pushes the `CreateWindow` CxOsOp, it may not be safe to always call this.
* are there any conditions in which you can call it safely but the passed-in values won't take effect as expected?

platform/src/window.rs
```diff
@@ -181,7 +181,16 @@ impl WindowHandle {
...
+        cx.windows[self.window_id()].create_title = title;
+        cx.windows[self.window_id()].create_position = Some(position);
+        cx.windows[self.window_id()].create_inner_size = Some(inner_size);
+        cx.windows[self.window_id()].is_created = true;
```
kevinaboos:
```suggestion
        let window = &cx.windows[self.window_id()];
        window.create_title = title;
        window.create_position = Some(position);
        window.create_inner_size = Some(inner_size);
        window.is_created = true;
```

kevinaboos (changes requested):

platform/src/window.rs
alanpoon: Changed to configure_window. The function only does its job in handle startup function.

kevinaboos (approved):

kevinaboos:
@alanpoon is this ready? it looks good to me; if so, please mark it as "ready for review" and i'll ping Rik about it.

PR #671, opened 2025-03-07
Add a generic spinner

emmettlu:
[spiner.webm](https://github.com/user-attachments/assets/e1e585ab-5e4d-4c13-b839-906e323d5d4f)

PR #774, opened 2025-06-23
widget: loading_spinner

Guocork:
Create a simple loading spinner using Sdf2d.

[录屏 2025-06-23 15-03-40.webm](https://github.com/user-attachments/assets/03ed584a-c942-4a19-8477-b543c6db84b5)

### DIFF
--- a/widgets/src/loading_spinner.rs
+++ b/widgets/src/loading_spinner.rs
@@ -6,58 +6,51 @@ live_design!{
     use link::widgets::View;
 
     pub LoadingSpinner = <View> {
-        width: Fill,
-        height: Fill
-        show_bg: true,
-        draw_bg: {
-            color: #00000000
-            
-            instance rotation_speed: 2.0
-            instance thickness: 3.0
-            instance radius: 30.0
-            instance arc_cycle_speed: 2.5
-            instance max_arc_ratio: 0.85
-            instance min_arc_ratio: 0.02
-            
-            fn pixel(self) -> vec4 {
-                let sdf = Sdf2d::viewport(self.pos * self.rect_size);
+            width: Fill,
+            height: Fill
+            show_bg: true,
+            draw_bg: {
+                color: #00000000
                 
-                let center = self.rect_size * 0.5;
+                instance rotation_speed: 1.2
+                instance stroke_width: 6.0
+                instance radius: 30.0
+                instance max_gap_ratio: 0.92
+                instance min_gap_ratio: 0.12
                 
-                sdf.circle(center.x, center.y, self.radius);
-                let stroke_color = vec4(1.0, 1.0, 1.0, 1.0);
-                sdf.stroke(stroke_color, self.thickness);
-                
-                let to_center = self.pos * self.rect_size - center;
-                let pixel_angle = atan(to_center.y, to_center.x);
-                
-                let rotation = self.time * self.rotation_speed * 2.0 * PI;
-                
-                let arc_cycle = sin(self.time * self.arc_cycle_speed) * 0.5 + 0.5;
-                let current_arc_ratio = mix(self.min_arc_ratio, self.max_arc_ratio, arc_cycle);
-                let current_arc_length = current_arc_ratio * 2.0 * PI;
-                
-                let arc_start = rotation;
-                let arc_end = arc_start + current_arc_length;
-                
-                let normalized_pixel_angle = mod(arc_start - pixel_angle + PI, 2.0 * PI);
-                let normalized_arc_length = mod(current_arc_length, 2.0 * PI);
-                
-                let in_arc = step(0.0, normalized_pixel_angle) * (1.0 - step(normalized_arc_length, normalized_pixel_angle));
-                
-                let fade_width = min(0.4, normalized_arc_length * 0.15);
-                let head_fade = smoothstep(0.0, fade_width, normalized_pixel_angle);
-                let tail_fade = smoothstep(normalized_arc_length - fade_width, normalized_arc_length, normalized_pixel_angle);
-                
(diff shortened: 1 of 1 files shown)


alanpoon:
I have tested on macOS and window.

kevinaboos:
There is already https://github.com/makepad/makepad/pull/671 merged in. Does that suit your purpose? If this is a better implementation, then kindly modify the existing loading spinner instead.

kevinaboos:
The prior animation from https://github.com/makepad/makepad/pull/671 looked better. Can you preserve the same appearance and style? (the rotation should be clockwise)

Guocork:
Hello, @kevinaboos , i have convert the style: 

https://github.com/user-attachments/assets/54e57add-7f35-4b33-9352-0d4bc45c5a35

kevinaboos:
Looks better!  The timing and arc length is a bit weird still. Here's what we want to emulate (the one on the right side): https://firebasestorage.googleapis.com/v0/b/design-spec/o/projects%2Fgoogle-material-3%2Fimages%2Flx1i8fz8-1P-progress-indicator-guidelines-types-determinate-indeterminate%202.mp4?alt=media&token=c408967f-7cd4-49db-8518-e032692dd6b0

kevinaboos (changes requested):
see above comment

Guocork:
hello @kevinaboos ,I'm currently modifying this component. I've created a new version using the `arc_round_caps` function (though I haven't committed the code yet). Here's the current result:

https://github.com/user-attachments/assets/0caf0e14-465c-413a-8bc4-2ce9f880fdc0

I set a specific color for the loading spinner to distinguish it from the white background in Robrix.May I ask if Makepad or Robrix has any specific requirements or guidelines for component colors?

kevinaboos:
That definitely looks better! Make sure to match the speed of the animation, and also the length of the arc (at its shortest point in particular). If you play them side-by-side, we shouldn't be able to notice any difference.

Guocork:
> That definitely looks better! Make sure to match the speed of the animation, and also the length of the arc (at its shortest point in particular). If you play them side-by-side, we shouldn't be able to notice any difference.

ok, i guess I can tweak the timing when the long arc transitions into the short arc.

Guocork:
hello, @kevinaboos I tweaked the rotation speed of the animation — this should feel more natural for users to look at.

https://github.com/user-attachments/assets/c09bc7bd-0780-4c4d-b3fe-04ea6aecb710

kevinaboos:
Perfect, nice work!

kevinaboos (approved):

widgets/src/loading_spinner.rs
```diff
@@ -6,68 +6,50 @@ live_design!{
...
         show_bg: true,
         draw_bg: {
-            color: #FFFFFF00
-            instance rotation_speed: 1.0
-            instance inner_radius: 0.6 // r
-            instance outer_radius: 0.7 // R
-
+            color: #00000000
+            
+            instance rotation_speed: 1.2
```
kevinaboos: I think all of these can be `uniform` instead of instances.

widgets/src/loading_spinner.rs
```diff
@@ -6,68 +6,50 @@ live_design!{
...
+                sdf.arc_round_caps(
+                    center.x, 
+                    center.y, 
+                    self.radius, 
+                    start_angle, 
+                    start_angle + 2.0 * PI - gap_radians, 
+                    self.stroke_width
+                );
+                
+                let spinner_color = vec3(0.29, 0.56, 0.89);
```
kevinaboos: Spinner color should be a uniform as well, such that others can set it. And you should hardcode the alpha channel to 1.0,  just pass in the color directly into `sdf.fill()`

kevinaboos (changes requested):
A few code comments

widgets/src/loading_spinner.rs
Guocork: done

widgets/src/loading_spinner.rs
Guocork: done

widgets/src/loading_spinner.rs
```diff
@@ -6,68 +6,49 @@ live_design!{
...
-            instance outer_radius: 0.7 // R
-
+            color: #00000000
+            
+            uniform rotation_speed: 1.2
+            uniform stroke_width: 6.0
+            uniform radius: 30.0
+            uniform max_gap_ratio: 0.92
+            uniform min_gap_ratio: 0.12
+            uniform spinner_color: vec3(0.29, 0.56, 0.89)
```
kevinaboos: this should just be a regular color constant, which itself will resolve to a `vec4`. 

I specifically said in the last PR that we don't want to force the alpha channel to be 1.0, but you're still doing that here.

```suggestion
            uniform spinner_color: #4A8FE3
```

widgets/src/loading_spinner.rs
```diff
@@ -6,68 +6,49 @@ live_design!{
...
+                sdf.arc_round_caps(
+                    center.x, 
+                    center.y, 
+                    self.radius, 
+                    start_angle, 
+                    start_angle + 2.0 * PI - gap_radians, 
+                    self.stroke_width
+                );
+                
+                return sdf.fill(vec4(self.spinner_color, 1.0));
```
kevinaboos: see my above comment.

```suggestion
                return sdf.fill(self.spinner_color);
```

kevinaboos (changes requested):

Guocork:
Got it, thanks for your reviewing

kevinaboos (approved):
Looks good, nice work!

PR #777, opened 2025-07-06
Add Size::Ratio

CN-MengXin:
Add a Ratio variant to Size for parent-relative proportional sizing.


eddybruel (changes requested):
This is a good start, in the sense that it correctly updates the implementation of eval_width and eval_height, but it is also incomplete, in the sense that it does not update the implementation of deferred walks.

To provide some context, we use deferred walks when you have situations like this: you walk the turtle three times, where the width of each walk is Size::Fill, Size::Fixed, and Size::Fill, respectively. In this case, the width that is not used by the Size::Fixed needs to be evenly distributed over the Size::Fills, but we cannot know what this unused width is when the first Size::Fill walk is performed.

So instead we defer the first walk, and resolve it when the turtle has finished (i.e. we call cx.end_turtle()). Resolving a deferred walk involves (among other things) shifting all the walks that come after a deferred walk to make space for the deferred walk now that we know its actual width.

Before this can land, we need to make sure that Size::Ratio will work with deferred walks. This is somewhat tricky, because right now the code involved with this assumes that although the size of the walk in the walk direction of the turtle (in the example above this was the width, but it can also be the height if the turtle walks down) cannot be known up front, the size of the walk in the other direction *can*. Once we introduce Size::Ratio the size of the walk in the other direction can no longer be known up front either, since it depends on the size of the walk in the direction of the turtle.

I should also warn you that I'm currently doing a big refactor of the turtle code, which among other things will add support for weighted and min/max fills, that is scheduled to land somewhere this week. If you felt so inclined to implement full support for Size::Ratio, I would recommend you hold off until this refactor lands at least, to avoid you having to rebase everything again later.

PR #782, opened 2025-07-11
fix_small_spinner_not_showing

alanpoon:
Loading Spinner is not showing when the width and height is 20.

Added logic to ensure the stroke does not exceed the viewport.


kevinaboos (approved):
thanks Alan!

PR #793, opened 2025-09-17
feat: a native wayland backend

drindr:


drindr:
I notice that in the code, Makepad seems to follow some non-explicit rules, like keeping the smallest dependence on other crates. 
And I'd like to introduce a crate from [wayland-rs](https://github.com/Smithay/wayland-rs), because it's absolutely more convenient for the coding experience.

So I'd like to comment here to ask if my observation is correct and if my solution is acceptable for Makepad's group. @okapii  @eddybruel

drindr:
Some small refactors and code clean works are still not complete.

PR #798, opened 2025-10-06
Fix macOS high CPU usage during idle

yulnr:
Addresses some regressions causing 10-20% idle CPU usage on macOS, reducing it to near 0%.

## Context
The macOS platform implementation had a timer (timer 0) running continuously every 8ms, even when the application had no work to do. This caused unnecessary CPU usage and prevented the RunLoop from properly sleeping.

### Symptoms
- ~10-20% CPU usage when idle on example-simple, and even with a completely empty window
- Taking samples of the stack showed continuous timer firing:
  ```
  __CFRunLoopDoTimers → received_timer → handle_repaint → will_redraw
  ```

### Causes

- **Unconditional Timer**: Timer 0 was started at startup (`start_timer(0, 0.008, true)`) and ran forever
- **Forced Repaints**: The `keep_alive_counter` mechanism called `repaint_windows()` which unconditionally marked all window passes as dirty, forcing unnecessary redraws
- **Always Polling EventFlow**: The event callback always returned `EventFlow::Poll` instead of `EventFlow::Wait`, preventing the RunLoop from blocking (there was some commented out code here that I restored)
- **File Watcher in Release**: Live reload file watcher thread was running even in release builds, polling every 100ms

### Dynamic Timer Management
- Timer 0 is no longer started unconditionally at startup
- Added `timer0_armed` field to track timer state
- Timer starts only when needed:
  - User activity (mouse/keyboard events)
  - Actual redraw work pending
  - Animation frames scheduled
- Timer stops when truly idle

### Fixed Repaint Logic
- Removed unnecessary `repaint_windows()` call in keep-alive counter handling
- Keep-alive now only maintains timer responsiveness, doesn't force repaints
- This eliminated the cycle of: timer → repaint_windows → dirty passes → more repaints

### Proper Event Flow Control
- Return `EventFlow::Wait` when idle (allows RunLoop to block)
- Return `EventFlow::Poll` only when work is pending
- This allows the macOS event loop to properly sleep using `distantFuture` instead of `distantPast`

### Conditional File Watcher
File watcher is now only compiled in debug builds, eliminating this thread polling saves additional 2-5% CPU in production

## Result
- **Idle CPU**: Reduced from ~10-20% to 0%
- **Main thread**: Properly blocks in `mach_msg2_trap` when idle
- **Timer behavior**: No timer events when idle (confirmed via stack sampling)
- **Mouse hover**: Still responsive but with some CPU usage due to keep-alive mechanism (future optimization opportunity)

## Future Improvements

For more efficient frame timing/handling we might want to integrate something like [CADisplayLink](https://developer.apple.com/documentation/quartzcore/cadisplaylink) alongside some form of invalidation  or pause during idle (similar to how Flutter does it).
This would also add support for vsync and OS-level adaptive frame rate, same for iOS with ProMotion displays (iPhone Pro).

### DIFF
--- a/platform/src/os/apple/macos/macos.rs
+++ b/platform/src/os/apple/macos/macos.rs
@@ -172,12 +172,13 @@ impl Cx {
                 event_flow
             }
         }));
-        // lets set our signal poll timer
-        // final bit of initflow
-        with_macos_app(|app| app.start_timer(0, 0.008, true));
-        
+
         cx.borrow_mut().call_event_handler(&Event::Startup);
         cx.borrow_mut().redraw_all();
+        // Start timer if there's initial work after startup
+        if cx.borrow().need_redrawing() {
+            cx.borrow_mut().ensure_timer0_started();
+        }
         MacosApp::event_loop();
     }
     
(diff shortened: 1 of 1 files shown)


makepaddev:
Ok sory or the late review, the reason timer 0 exists is it 'polls' the multithreaded signals. Without that you have no way to send a message from a thread in a timely manner, unless there is still something running

makepaddev:
I'll merge it in since obviously running hot on CPU is bad, but i need to investigate how thread events will now work

PR #803, opened 2025-10-24
Fix button clicks during keyboard dismissal on mobile

yulnr:
### Problem
On mobile, tapping a button while the soft keyboard is visible causes:
1. Show the button's pressed animation (correct)
2. Dismiss the keyboard (correct)
3. **Fail to register the click (incorrect)**

This requires a second tap from the user on the button to actually trigger the action. 
Note that this only happens if the button position was shifted by the displaying of the soft keyboard (noticed mostly in Robrix and Moly).

### Root Cause
The sequence of events:
1. User taps button -> `FingerDown` event -> button shows pressed state
2. Makepad detects tap outside text field -> dismisses keyboard immediately
3. `KeyboardView` animates layout back down -> **button moves** (let's say ~500px)
4. User lifts finger -> `FingerUp` event
5. Event system checks: is finger still over button? -> **no** (button moved)
6. Click not registered 

The finger system is calculating `is_over` based solely on whether the finger position was inside the widget's rect at `FingerUp` time, not accounting for layout shifts that occur mid-press.

### Solution

Extended `is_over` calculation to handle two cases:
1. **Normal:** Finger is within widget's current rect
2. **Layout shift:** Finger didn't move significantly from start position (within 5px and <0.5s)

```rust
let is_over = rect.contains(t.abs) || (
    (e.time - capture.time < TAP_COUNT_TIME) &&
    ((t.abs - capture.abs_start).length() < TAP_COUNT_DISTANCE)
);


makepaddev:
This now conflicts with your touch radius changes

yulnr:
@makepaddev fixed! should be ready to go now
sorry for the PR spam, could've included them in one

PR #839, opened 2026-01-16
IME: Support on Android and Input Configuration

yulnr:
## IME & Text Input: Native Soft Keyboard Support (iOS + Android)

Implements native IME (Input Method Editor) support for iOS and Android, enabling proper soft keyboard integration with composition preview, autocorrect, and configurable keyboard types.

### Cross-Platform API

New `TextInputConfig` struct for unified keyboard configuration:

- `InputMode`: Text, Ascii, Url, Numeric, Tel, Email, Decimal, Search (matches web `inputmode` attribute)
- `AutoCapitalize`: None, Words, Sentences, AllCharacters
- `AutoCorrect`: Default, Enabled, Disabled
- `ReturnKeyType`: Default, Go, Search, Send, Done

```
<TextInput> {
    empty_text: "https://example.com"
    is_multiline: false
    input_mode: Url
    autocorrect: Disabled
    autocapitalize: None
    return_key_type: Go
}
```

### What Changed

**Platform layer (`platform/src/`)**
- **iOS**: Replaces the previous hidden UITextField approach with a full `UITextInput` protocol implementation. Supports marked text (composition) for CJK input methods, autocorrect via `replaceRange:withText:`, keyboard show/hide animations, and `UIEditMenuInteraction` for copy/paste/cut (iOS 16+). CJK-aware autocorrection (auto-disables for Chinese/Japanese/Korean keyboards).
- **Android**: Implements `MakepadInputConnection` in Java for proper IME integration. Full-state sync model where Java's `mEditable` is the source of truth. Handles composition spans, editor actions, batch edits, and input filtering. Floating `ActionMode` for clipboard operations. Proper `EditorInfo` configuration for all input modes.
- **Shared**: `TextInputConfig` API modeled after web `inputmode`/`autocapitalize` attributes. `SyncImeState` platform op for programmatic text updates (clear, paste, undo). `CharOffset` type for platform-independent text position communication.

**Widget layer (`widgets/src/text_input.rs`)**
- `Hit::TextInput` handler supporting both Android's full-state sync and iOS's incremental range-replace model.
- Composition underline rendering for active IME previews.
- `filter_input()` enforces `InputMode` constraints (ASCII, Numeric, Decimal, Tel) on all platforms.
- Clipboard actions (long-press menu) on mobile with proper keyboard-shift awareness.
- Tap-to-position, double-tap-to-select-word, triple-tap-to-select-all on touch.

**Example app (`examples/text_input/`)**
- New `makepad-example-text-input` app demonstrating all input configurations.

### Architecture

The sync model differs per platform because of OS constraints:

| | Android | iOS | macOS/Desktop |
|---|---|---|---|
| Source of truth | Java `mEditable` | Rust widget | Rust widget |
| Sync direction | Bidirectional | iOS -> Rust | Events only |
| Composition | Composing spans in `mEditable` | `NSMutableAttributedString` ivar | `NSTextInputClient` |
| Echo prevention | `mLastSentText` + widget-side state cache | Not needed (unidirectional) | Not needed |
| Keyboard config | `EditorInfo` flags | UITextInputView ivars | Ignored (no soft keyboard) |

Android requires a Java-side shadow buffer because `InputConnection` must answer synchronous queries (`getTextBeforeCursor`, etc.). iOS's `UITextInput` is more event-driven.

`TextInputConfig` properties (`autocapitalize`, `autocorrect`, `return_key_type`) are mobile soft-keyboard hints only -- no effect on desktop. `InputMode` filtering is enforced at the widget layer on all platforms.

### Breaking Changes

- `show_clipboard_actions()` signature changed to include positioning info.
- `is_numeric_only` removed from TextInput -- use `input_mode: Decimal` instead.

### Known Limitations

- **`selectionRectsForRange:`** (iOS): Returns empty. Would need layout info from the widget layer. System selection handles won't show exact rects, but input is functional.
- **Large documents**: Android sends full text on every change. Fine for typical inputs, not suited for large documents (code_editor has its own input path).
- **Focus transitions**: Composition state may persist incorrectly across focus changes.

### Not Implemented (Future PRs)

- RTL text (Arabic, Hebrew)
- Hardware keyboard modifiers on mobile
- Android autofill framework
- `ReturnKeyType::Next` field navigation
- iPad split keyboard candidate window positioning

### Testing

- `cargo build -p makepad-widgets` (desktop)
- `cargo build -p makepad-platform --target aarch64-apple-ios`
- Manual testing: CJK input (Japanese, Chinese), emoji, autocorrect, keyboard types, single-line/multiline, password fields
- Tested on iOS native keyboard, Android GBoard, Swift Keyboard, Samsung Keyboard
- Example app: `cargo makepad android run -p makepad-example-text-input`


tools/cargo_makepad/src/apple/compile.rs
```diff
@@ -790,22 +790,28 @@ pub fn run_on_device(apple_args: AppleArgs, args: &[String], apple_target: Apple
...
+                "device",
+                "process",
+                "launch",
+                "--device",
+                device_identifier,
+                &bundle_id
+            ])?;
+        } else {
+            return Err(format!("Failed to find bundleID in installation output"));
+        }
```
yulnr, line 814: Unrelated fix for iOS device launch: `devicectl device process launch` expects a bundle identifier (e.g., com.org.app), not a file path. The old code parsed `installationURL:` with a hardcoded string offset. Now properly extracts `bundleID:` from the installation output.

widgets/src/text_input.rs
```diff
@@ -614,10 +639,18 @@ pub struct TextInput {
     #[rust] preserved_selection_cursor: Option<Cursor>,
     /// Skip finger move after long press to prevent selection changes
     #[rust] ignore_next_move: bool,
-    /// IME composition tracking - byte index where composition starts
+    /// IME composition tracking - byte index in self.text where composition starts
     #[rust] composition_start: usize,
-    /// IME composition tracking - byte length of current composition
+    /// IME composition tracking - byte length of current composition in self.text
     #[rust] composition_length: usize,
+    /// Frame ID when IME input was last received; prevents race condition in update_ime_context
```
ejpbruel2: Putting a comment here that this exists to prevent a race condition is great! Can you also add an explanation how this race condition occurs and how this field prevents it? That would make this comment even more helpful.

widgets/src/text_input.rs
```diff
@@ -614,10 +639,18 @@ pub struct TextInput {
...
+    /// Uses frame counter instead of bool to avoid timing issues between event/draw
+    /// If this matches cx.redraw_id(), we skip syncing to avoid echoing state back.
+    #[rust] ime_update_frame: u64,
+    /// Track last sent IME state to prevent sync loops
```
ejpbruel2: As someone who is not familiar with IME: what are sync loops and why do they occur? Explaining that here would make this comment even more helpful.

widgets/src/text_input.rs
```diff
@@ -666,18 +699,16 @@ impl TextInput {
...
-    pub fn set_is_numeric_only(&mut self, cx: &mut Cx, is_numeric_only: bool) {
-        self.is_numeric_only = is_numeric_only;
-        self.laidout_text = None;
-        self.draw_bg.redraw(cx);
-    }
-
-    pub fn toggle_is_numeric_only(&mut self, cx: &mut Cx) {
-        self.set_is_numeric_only(cx, !self.is_numeric_only);
+    /// Build IME configuration from widget properties for mobile platforms
+    pub fn get_ime_config(&self) -> TextInputConfig {
```
ejpbruel2, line 758: If this config is purely meant for IME, wouldn't TextInputIMEConfig be a better name?

widgets/src/text_input.rs
```diff
@@ -723,6 +754,45 @@ impl TextInput {
...
+            || sel_end_utf16 != self.last_sent_ime_sel_end
+        {
+            self.last_sent_ime_text = self.text.clone();
+            self.last_sent_ime_sel_start = sel_start_utf16;
+            self.last_sent_ime_sel_end = sel_end_utf16;
+
+            // Sync via unified operation
+            cx.sync_ime_state(
+                self.text.clone(),
+                CharOffset(sel_start_chars)..CharOffset(sel_end_chars),
```
ejpbruel2, line 848: Is there a strong reason to prefert char offsets over byte offsets here? Char offsets are typically inefficient because you have to walk over the entire string to either use them as index or compute them (as you can see in your computation of sel_start_chars and sel_end_chars above)

widgets/src/text_input.rs
```diff
@@ -912,6 +982,41 @@ impl TextInput {
...
+        if self.composition_length == 0 {
+            return;
+        }
+
+        let laidout_text = self
+            .laidout_text
+            .as_ref()
+            .expect("layout should never be `None` here");
+
+        let comp_end = (self.composition_start + self.composition_length).min(self.text.len());
```
ejpbruel2: Small nit: why not composition_end?

widgets/src/text_input.rs
```diff
@@ -1691,49 +1866,132 @@ impl Widget for TextInput {
...
-            Hit::TextInput(TextInputEvent {
-                input,
-                replace_last,
-                was_paste,
-                ..
-            }) if !self.is_read_only => {
-                let input = self.filter_input(&input, false);
+            Hit::TextInput(event) if !self.is_read_only => {
+                // Text changes invalidate any preserved cursor from a pending tap gesture
+                self.preserved_selection_cursor = None;
```
ejpbruel2, line 1928: I don't think I've seen this field before, and there's no comment explaining it. What is preserved_selection_cursor for?

widgets/src/text_input.rs
```diff
@@ -723,6 +754,45 @@ impl TextInput {
...
+        }
+
+        use crate::makepad_platform::event::keyboard::CharOffset;
+
+        // Convert byte indices to character offsets
+        let sel_start_chars = self.text[..self.selection.start().index].chars().count();
+        let sel_end_chars = self.text[..self.selection.end().index].chars().count();
+
+        // Convert to UTF-16 for comparison (Android uses UTF-16)
+        let sel_start_utf16 = byte_index_to_utf16_index(&self.text, self.selection.start().index);
```
ejpbruel2: Probably does not matter in practice, but this is needlessly inefficient when you have a selection near the end of a large text. Both calls to byte_index_to_utf16_index iterate over the entire string. One improvement would be to do the second call to byte_index_to_utf16_index on &self.text[self.selection.start().index..] and then add the result to sel_start_utf16 to obtain sel_end_utf16.

widgets/src/text_input.rs
```diff
@@ -723,6 +754,45 @@ impl TextInput {
...
+        let sel_end_utf16 = byte_index_to_utf16_index(&self.text, self.selection.end().index);
+
+        // Only send if state actually changed from what we last sent
+        // This prevents the sync loop where IME sends state → we echo it back → IME gets confused
+        if self.text != self.last_sent_ime_text
```
ejpbruel2, line 837: Can this still happen even though we only update the ime state when the ime frame matches the draw frame?

widgets/src/text_input.rs
```diff
@@ -1691,49 +1866,132 @@ impl Widget for TextInput {
...
+                        self.composition_start = comp_start_byte;
+                        self.composition_length = comp_end_byte - comp_start_byte;
+                    } else {
+                        self.composition_start = 0;
+                        self.composition_length = 0;
+                    }
+
+                    // Track sent state to prevent sync loops
+                    self.last_sent_ime_text = self.text.clone();
+                    self.last_sent_ime_sel_start = full_state.selection.start.to_utf16_index(&self.text);
```
ejpbruel2: You do work here to convert the start of the selection to from an utf8 index to an utf16 index, before storing it in last_sent_ime_sel_start. Later on, on line 774, you again do work to convert the start of the selection to an utf16 index, just so you can compare it to last_sent_ime_sel_start. Would it not be more efficient to just store the index as an utf8 index?

widgets/src/text_input.rs
```diff
@@ -1691,49 +1866,132 @@ impl Widget for TextInput {
...
+
+                    self.draw_bg.redraw(cx);
+                    cx.widget_action(uid, &scope.path, TextInputAction::Changed(self.text.clone()));
+                    if text_changed {
+                        cx.hide_clipboard_actions();
+                    }
+                    return;
+                }
+
+                // Handle iOS range replacement (autocorrect/paste)
```
ejpbruel2, line 1970: Why does iOS range replacement need special handling? Some clarification here would be helpful.

widgets/src/text_input.rs
```diff
@@ -1782,36 +2043,27 @@ impl Widget for TextInput {
...
+                        cx.revert_key_focus();
+                        cx.widget_action(uid, &scope.path, TextInputAction::Returned(self.text.clone(), mods));
                     }
-                );
-
-                self.animator_play(cx, ids!(empty.off));
-                self.draw_bg.redraw(cx);
-                cx.widget_action(uid, &scope.path, TextInputAction::Changed(self.text.clone()));
+                    ImeAction::Next | ImeAction::Previous => {
+                        cx.widget_action(uid, &scope.path, TextInputAction::Returned(self.text.clone(), mods));
```
ejpbruel2, line 2129: Why does this yield a TextInputAction::Returned? A comment here would be helpful

widgets/src/text_input.rs
```diff
@@ -1782,36 +2043,27 @@ impl Widget for TextInput {
                 self.animator_play(cx, ids!(empty.off));
                 self.draw_bg.redraw(cx);
                 cx.widget_action(uid, &scope.path, TextInputAction::Changed(self.text.clone()));
+                if self.ime_update_frame != cx.redraw_id() {
+                    self.update_ime_context(cx);
```
ejpbruel2: Why do we update the ime context here if this is already done on draw?

widgets/src/text_input.rs
```diff
@@ -1746,24 +2004,27 @@ impl Widget for TextInput {
...
                     if self.composition_length > 0 {
                         // Replace composition with final committed text
+                        let start = self.composition_start.min(self.text.len());
+                        let end = (self.composition_start + self.composition_length).min(self.text.len());
                         self.create_or_extend_edit_group(EditKind::Other);
                         self.apply_edit(
                             cx,
                             Edit {
-                                start: self.composition_start,
-                                end: self.composition_start + self.composition_length,
```
ejpbruel2, line 1758: I've noticed that mosf of the time you really need self.composition_end and you compute it from self.composition_start and self.composition_length. Would it not be more convenient to just store composition_end? (Also more consistent with the rest of the code, which tends to use start/end for things like selections)

widgets/src/text_input.rs
```diff
@@ -1691,49 +1866,132 @@ impl Widget for TextInput {
...
+                        cx.hide_clipboard_actions();
+                    }
+                    return;
+                }
+
+                // Handle iOS range replacement (autocorrect/paste)
+                if let Some((start, end)) = event.replace_range {
+                    let filtered_text = self.filter_input(&event.input, false);
+                    if filtered_text.is_empty() && !event.input.is_empty() {
+                        // Re-sync IME state to reject invalid input
```
ejpbruel2: Is input invalid if ALL characters are filtered? What if only SOME characters are filtered?

widgets/src/text_input.rs
```diff
@@ -1691,49 +1866,132 @@ impl Widget for TextInput {
...
+                        self.update_ime_context(cx);
+                        return;
+                    }
+
+                    // Convert character offsets to byte indices
+                    let byte_start = start.to_byte_index(&self.text);
```
ejpbruel2, line 1987: Seeing this, I'm guessing the reason you use CharOffsets here and elsewhere is that because that's just how Makepad's event API is designed. I still think that's wrong, but not your responsibility to fix then :-)

ejpbruel2 (approved):

widgets/src/text_input.rs
yulnr: Good catch, not needed here anymore

widgets/src/text_input.rs
yulnr: Good catch, fixed.

widgets/src/text_input.rs
yulnr: Good idea. 
I don't know why initially I assumed this as default behavior, but after your comment I checked the web and some UI frameworks and it seems like in general a numeric-only input will allow you to paste a "abc123" and take "123", rather than invalidate the input entirely. I updated to match that.

widgets/src/text_input.rs
yulnr: Yup, I didn't dare to rework much more in this PR, but it's a good point we should consider soon

widgets/src/text_input.rs
yulnr: Good idea, since `multi_line` and `is_secure` are also used outside of IME, I'm refactoring into a more general `TextInputConfig`, that also has soft keyboard specific settings:
```rust
/// Text input configuration combining cross-platform and mobile-specific settings.
#[derive(Clone, Copy, Debug, Default, PartialEq)]
pub struct TextInputConfig {
    /// Soft keyboard settings for mobile platforms. No effect on desktop.
    pub soft_keyboard: SoftKeyboardConfig,
    /// Whether the input field supports multiple lines.
    pub is_multiline: bool,
    /// Whether to mask input characters (password field).
    pub is_secure: bool,
}
```

yulnr:
Hi @ejpbruel2 thanks for thorough review!
You caught some unnecessary IME context syncing calls, and some inefficiencies that had some easy fixes so I've applied them.
I've also added better comments and documentation that I'm hoping suffices. I've moved most things that I added into `platform/cx_api.rs` to its own module `platform/ime.rs` and I left there some general documentation of how IME works (and differs between platforms) in Makepad. But hopefully the comments and docs on `TextInput` explain enough why things are being done. 

if you still find something here (now are anytime in the future) that could be improved (be it code or docs) just let me know :)

I've also replied to some of your comments above, mostly just agreeing with you so I resolved them.

PR #842, opened 2026-01-19
FoldHeader hides the body when opened < 0.1

alanpoon:
FolderHeader current does not hide the body when it is closed. Causing a big gap with the next item after it is closed.
https://github.com/user-attachments/assets/9a839294-669b-45c3-8b78-a88a77da74bf

### DIFF
--- a/widgets/src/fold_header.rs
+++ b/widgets/src/fold_header.rs
@@ -71,7 +71,9 @@ impl Widget for FoldHeader {
         };
         
         self.header.handle_event(cx,  event, scope);
-        
+        if self.body.visible() {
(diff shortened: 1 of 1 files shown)


alanpoon:
> FolderHeader current does not hide the body when it is closed. Causing a big gap with the next item after it is closed. https://github.com/user-attachments/assets/9a839294-669b-45c3-8b78-a88a77da74bf

https://github.com/project-robius/robrix/pull/637


https://github.com/user-attachments/assets/2c3d5a0e-f452-49ef-968a-b05a4c469884

PR #850, opened 2026-01-31
Allow customize binary name in the wasm build tool

drindr:
Sometime, the binary name is different from the crate name. This PR makes it available with the makepad-cargo tools.


offline-ant:
fwiw this commit https://github.com/makepad/makepad/commit/31f2a410387a389151db74cc37c8fb620cbc9f7f changed some things and reads the bin name from json to create mac/android packages that use the bin name. It should be straightforward to have wasm to do the same.

PR #864, opened 2026-02-21
linux: wayland, stdin routing, control channel, studio crash fix, Y-flip fix

offline-ant:
Linux and platform improvements:

- **wayland**: edge-resize, scroll events (wheel/touchpad distinction), wl_seat v5 binding, time-based scroll acceleration for discrete wheel
- **stdin bugfix**: route all StudioToApp variants through stdin_handle_host_to_stdin — fixes dead match arms where Screenshot, WidgetTreeDump, Kill were unreachable (linux + windows)
- **widgets**: expose Label and View fields for external composition
- **platform**: stdin/stdout control channel for windowed apps — lets windowed apps receive StudioToApp messages via a channel and emit AppToStudio on stdout. Includes GL readback for windowed-mode screenshots.
- **unified dispatch**: extract shared dispatch_studio_msg() in cx_shared.rs so both stdin mode and the control channel use the same StudioToApp handler. Eliminates ~130 lines of duplicated dispatch logic. Control channel polling now hooked into all backends (X11, Wayland, macOS).
- **atan2 fix**: map atan2 to atan in GLSL shader backend
- **WindowGeomChange**: stdin emits WindowGeomChange event on geometry updates
- **aux-channel**: defer aux-channel accept until after build finishes — prevents race where Studio connects before the app is ready
- **Y-flip fix**: fix RunView Y-flip on Linux by flipping in shader instead of CPU readback
- **Studio crash fix**: fix Studio crash on child process failure

### DIFF
--- a/platform/script/src/shader_backend.rs
+++ b/platform/script/src/shader_backend.rs
@@ -1144,6 +1144,7 @@ impl ShaderBackend {
                     // GLSL uses dFdx/dFdy natively, mod is native
                     id!(inverseSqrt) => id!(inversesqrt),
                     id!(modf) => id!(mod),
+                    id!(atan2) => id!(atan),
                     x => x,
                 }
             }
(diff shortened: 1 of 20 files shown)


offline-ant:
Closing — rebased and reorganized on top of latest upstream. Will reopen with updated PR.

PR #909, opened 2026-03-02
widget_tree: fix compact_dump scale field

offline-ant:
The O line's scale field was 1000 (a leftover from when coordinates were in physical pixels and scale encoded `dpi * 1000`). Coordinates are now logical, so scale should be 1.


makepaddev:
removed it from the protocol instead

PR #911, opened 2026-03-04
platform: popup/context menu window API

offline-ant:
Add cross-platform popup window operations to the platform API. Popup windows support parent-child relationships, positioning, and automatic dismissal on focus loss, outside click, or escape.

Platforms wired: X11, Wayland, macOS, Windows, iOS, Android. Stub-only on web, tvOS, OpenHarmony.

Validated with cargo check -p makepad-platform.


makepaddev:
we have merge conflicts on this one and its hard to see how to merge it correctly without a  linux testbox

PR #917, opened 2026-03-05
fix(draw): prevent panics when font data unavailable on WASM

dalbrecht:
## Summary

Fixes `RuntimeError: unreachable` panics on WASM when font binary data is not yet available via the JS dependency table during early frames.

### Root cause
On WASM, font resources arrive asynchronously via `ToWasmInit`. If `cx.get_resource(handle)` returns `None` for a font, `update_font_definitions()` in `draw_text.rs` silently skips it but still registers the font family. When the text system later attempts to load the family, the loader panics on `.expect()` / `.unwrap()`. This corrupts Makepad's event handler state, causing every subsequent frame to also panic.

### Changes

**`draw/src/text/loader.rs`** — 5 methods return `Option` instead of panicking:
- `load_font()` → `Option<Font>` (font definition or face parse failure returns `None`)
- `get_or_load_font()` → `Option<&Rc<Font>>`
- `load_font_family()` → `Option<FontFamily>` (skips failed fonts, returns `None` if all fail)
- `get_or_load_font_family()` → `Option<&Rc<FontFamily>>`
- `get_or_load_font_family_rc()` → `Option<Rc<FontFamily>>`

**`draw/src/text/layouter.rs`** — 2 methods:
- `get_or_load_font_family()` → `Option<Rc<FontFamily>>`
- `layout()` → returns empty `LaidoutText` when font family unavailable

**`draw/src/text/fonts.rs`** — 1 method:
- `get_or_load_font_family()` → `Option<Rc<FontFamily>>`

**`draw/src/shader/draw_text.rs`** — 2 methods:
- `draw_walk_laidout()` → early return with `Rect::default()` when rows are empty
- `draw_walk_resumable_with()` → early return when rows are empty

**`widgets/src/math_view.rs`** — 1 caller adapted to use `and_then`

### Testing
- Desktop build compiles and passes tests
- WASM build loads with **zero** `RuntimeError` panics (previously ~70 per-frame panics)
- Fonts render correctly once data becomes available

### DIFF
--- a/draw/src/shader/draw_text.rs
+++ b/draw/src/shader/draw_text.rs
@@ -390,6 +390,10 @@ impl DrawText {
         use crate::text::geom::{Point, Size};
         use crate::turtle;
 
+        if laidout_text.rows.is_empty() {
+            return Rect::default();
+        }
+
         let size_in_lpxs = laidout_text.size_in_lpxs * self.font_scale;
         let max_size_in_lpxs = Size::new(
             cx.turtle()
@@ -460,6 +464,9 @@ impl DrawText {
             Align::default(),
             text_str,
         );
+        if text.rows.is_empty() {
+            return;
+        }
         self.draw_text(cx, origin_in_lpxs, &text);
 
         let last_row = text.rows.last().unwrap();
--- a/draw/src/text/fonts.rs
+++ b/draw/src/text/fonts.rs
@@ -137,7 +137,7 @@ impl Fonts {
         self.layouter.define_font(id, definition);
     }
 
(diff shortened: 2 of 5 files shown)


makepaddev:
i'm all for merging fixes but i think you ran cargo fmt or something on the entire repo making this a 3000 file change.

dalbrecht:
> i'm all for merging fixes but i think you ran cargo fmt or something on the entire repo making this a 3000 file change.

Apologies, this PR had been pointed at the wrong branch.  It should only be changes to 5 files.

PR #919, opened 2026-03-05
Remove debug title update in redraw handler

wheregmis:
Remove a leftover debug call that set the document title when handling ToWasmRedrawAll in platform/src/os/web/web.rs. This avoids an unnecessary DOM update used only for debugging and cleans up the redraw handler.


wheregmis:
image
This fixes the title, which is shown based on windows title

PR #936, opened 2026-03-09
Hotreloading Nits

wheregmis:


makepaddev:
what does this do?

wheregmis:
Nothing huge, just bring the common implementation from platform and cargo makepad for native and wasm as seperate lib and using it on both platform and cargo makepad. (+ few tests)

PR #951, opened 2026-03-14
Fix PulseAudio input stream dedupe

offline-ant:
## Summary
- fix PulseAudio input stream dedupe to keep distinct input devices
- fix PulseAudio input callback userdata cast
- fix PulseAudio threaded-mainloop wait and signal handling for stream startup and device queries

## Testing
- cargo check -p makepad-platform


offline-ant:
Updated this PR with the threaded-mainloop wait/signal fix and reran `cargo check -p makepad-platform`.

PR #954, opened 2026-03-14
Fix macos platform build failures related to PR#946

kevinaboos:


kevinaboos:
Seems like #957 was chosen instead

PR #955, opened 2026-03-14
Avoid double refcell borrow in `Widget::area()` fn

kevinaboos:
This was causing loads of random runtime panics in Robrix. I'm not sure if this is the root cause but it is a fix at least.


kevinaboos:
PR #956 is probably a better solution here, but i'm not 100% sure yet so i'm marking this as a draft.

kevinaboos:
Closing in favor of #956

PR #964, opened 2026-03-17
Refine block/list spacing and plain-bullet handling

### LINKED ISSUE #959: Markdown ordered-list item marker renders on its own line
## Summary
When rendering Markdown with the Makepad `Markdown` widget, an ordered-list item can be laid out incorrectly: the numeric marker (for example `1.`) is rendered on its own line, while the item text starts on the next line.

This was observed in Moly when converting Telegram HTML into Markdown, but the resulting Markdown string is valid on its own and the problem appears to be in the upstream Makepad Markdown rendering/layout behavior.

## Minimal repro
Render this Markdown with the Makepad `Markdown` widget:

```md
1. **OpenAI Sora 2 API重大更新**
• 角色一致性大幅提升
• 支持20秒视频生成
```

## Actual behavior
The ordered-list marker (`1.`) is shown on its own line, and the bold heading text is rendered on the next line.

## Expected behavior
The marker and the first line of the list item should stay on the same line:

- `1. OpenAI Sora 2 API重大更新`

followed by the later lines below it.

## Context
- Observed in Moly: Telegram HTML like `1. <b>OpenAI Sora 2 API重大更新</b>` is normalized into Markdown before rendering.
- The downstream workaround is to escape ordered-list markers for this Telegram-specific path, but that is only an application-level compatibility fix.
- Current Makepad dependency in Moly is based on the Makepad code at commit `1aa19396c` (via `wyeworks/makepad`, but the issue appears to apply to upstream `makepad/makepad` Markdown behavior).

## Notes
This seems related to how ordered-list items are laid out when the first line is followed by additional block-like content. If helpful, I can provide a smaller standalone Makepad reproduction as well.

wheregmis:
Replace single is_first_block flag with a stack to track first-in-block state for nested containers (blockquotes, lists, items) and adjust spacing behavior accordingly. Add helper methods (enter_block, mark_current_container_non_first, text_starts_with_plain_bullet, next_text_starts_with_plain_bullet), collect parser events into a Vec for lookahead, and track list_item_depth to avoid inserting unwanted spaces/newlines before plain bullet markers in list items. Also tweak handling around headings, paragraphs, code/display math, images and links to correctly mark containers as non-first when inline content appears.

Fixes: #959


wheregmis:
Before:
Screenshot 2026-03-16 at 11 42 11 PM
After:
Screenshot 2026-03-16 at 11 42 43 PM

PR #968, opened 2026-03-18
Fix for portal list jumping to the top when scrolling upwards

kevinaboos:


kevinaboos:
actually i'm not sure if this is even needed. It introduces additional visual scrolling artifacts, so this probably isn't the best solution.

kevinaboos:
I don't think this is actually needed.

PR #985, opened 2026-03-25
Add running-builds UI and refactor HTTP server bind

wheregmis:
Introduce stop-builds support in the desktop run list UI and refactor HTTP server binding. The run list now shows running state, a stop icon/badge, groups running builds by package, and sends a StopBuilds action; DesktopRunListRef.requested_action and handlers in main/app_tabs/app_backend/messages were updated to handle the new action and refresh the run list. Added bind_http_server that returns the actual bound SocketAddr plus the thread handle and kept start_http_server as a convenience wrapper; gateway now uses bind_http_server to obtain the listen address. process_manager now computes a use_stdio_bridge flag and passes the correct stdio flag into start_command_run. Minor UI and test tweaks: adjusted sidebar input padding, refreshed visibility for run_stop_all, and disabled in-process gateway in headless tests.

### DIFF
--- a/libs/makepad_test/src/runtime.rs
+++ b/libs/makepad_test/src/runtime.rs
@@ -1152,7 +1152,7 @@ fn start_headless_app(config: &TestConfig) -> TestResult<(TestConnection, QueryI
                 name: config.mount_name.clone(),
                 path: config.manifest_dir.clone(),
             }],
-            enable_in_process_gateway: true,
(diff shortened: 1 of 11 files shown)


wheregmis:
Additionally,

When something is running
Screenshot 2026-03-25 at 2 46 59 PM

When nothing is runinng:
Screenshot 2026-03-25 at 2 47 17 PM

PR #996, opened 2026-03-27
platform: unify wake handling across macOS, iOS, Linux, and Windows

offline-ant:
Current local `dev` split, rebased onto `origin/dev`.

This branch intentionally includes the single-instance commit because the wake-handling changes sit on top of it in current local history.

Commits:
- 0c139851 platform: add single-instance AppOpen delivery
- 2f57c257 platform: unify wake handling across macOS, iOS, Linux, and Windows

Validation:
- `cargo check`


offline-ant:
Was not tested on windows

PR #997, opened 2026-03-27
Remove dead dependency system, add `ImageRef::load_image_from_resource_abs_path`

wyenox:
## Summary

The old `cx.dependencies` system (`CxDependency` HashMap, `*_load_dependencies()` loaders, `load_image_dep_by_path`) is dead code in the new script/splash system. Resources are now stored in `cx.script_data.resources` (populated by `crate_resource()`), and `cx.dependencies` is never populated on any platform — making every function that reads from it a no-op or guaranteed failure.

This PR removes all the dead code and adds a new `ImageRef::load_image_from_resource_abs_path` method to fill the gap left by the removed `load_image_dep_by_path`.

## Changes

**Removed (dead code cleanup):**
- `CxDependency` struct and `cx.dependencies` HashMap field
- `native_load_dependencies()`, `apple_bundle_load_dependencies()`, `android_load_dependencies()`, `ohos_load_dependencies()` — all iterated the empty HashMap
- `load_image_dep_by_path` from `ImageCacheImpl` trait, `ImageRef`, and `ImageBlendRef`
- `live_scan_dependencies()` calls in `linux_direct` and `open_harmony` (function no longer exists)
- All `*_load_dependencies()` call sites across platform init code (macos, ios, tvos, windows, linux, headless, ohos)

**Preserved:**
- `take_dependency()` / `get_dependency()` — stripped the dead HashMap lookup but kept the Android JNI (`to_java_load_asset`) fallback which is actively used by `load_script_resource` in `res.rs`
- `apple_bundle_load_file()` — used by the script resource system on iOS/tvOS

**Added:**
- `ImageRef::load_image_from_resource_abs_path(cx, abs_path)` — loads an image from the script resource system by the absolute path registered via `crate_resource()`. This is the replacement for the old `load_image_dep_by_path` for Rust code that needs to load images at runtime from paths not known at DSL time.


kevinaboos:
I've fixed packaging logic now in Makepad 2.0, so these functions are used again. I believe this can be closed now, or at least changed to only remove the dead code that isn't actually used.

PR #1000, opened 2026-03-31
Choose sane defaults for platform-specific title/caption bar config

kevinaboos:
Primarily on Linux, ensure that we show the title/caption bar and draw it within Makepad (i.e., client-side drawing) if the DE/WM doesn't show it by default.
This should make things behave as expected on Linux X11 and Wayland both.


kevinaboos:
verified working on Ubuntu 25.10 with Wayland.

kevinaboos:
verified working on Ubuntu 20.04 with X11. Ready to merge.

PR #1001, opened 2026-03-31
Expose window chrome button bounding box in `WindowGeom`

kevinaboos:
This allows apps that wanna draw something in the title/caption bar to do so in a proper way without potentially drawing over the native window chrome buttons (on macOS, the traffic light buttons). Without this it'd be pretty tough to figure that out.

This also auto-sets the caption bar height to be tall enough such that the window chrome / traffic light buttons are perfectly vertically-centered in the middle of the caption bar. This was needed on macOS to prevent things from looking janky as hell on newer macOS versions, which changed the default size of the traffic chrome buttons.
It'll also be useful for drawing things in the caption bar on linux or windows too. For example, with this you can fix the way that the show/hide sidebar buttons in Makepad Studio are padded on the left & right side of the caption bar.

If you want to manually override the caption bar's height, you can now do so with the window's property `override_caption_bar_height`, but there shouldn't really ever be a reason to do that in most normal apps.

I also fixed the draggable window-moving region to match that of the title bar, isntead of being an arbitrary 25x250 pixel rect.

-----------

Full change set:

- `widgets/src/window.rs`: Hide the caption bar on `LinuxWindow` when `!custom_window_chrome` (X11 — WM provides native decorations) directly in `sync_caption_bar_state`, removing the need for apps to do this manually.
- `event/window.rs`: Add `window_chrome_buttons: Rect` to `WindowGeom` — the bounding box of the OS/app-drawn window chrome buttons in logical window coordinates (top-left origin). Non-zero on macOS (traffic lights), Windows (min/max/close), and Wayland with `custom_window_chrome`. Zero on all other platforms (X11, LinuxDirect, mobile, web). Documented with per-platform details and guidance on how to use it for caption-bar layout margins.
- `macos_window.rs`: Add `traffic_lights_geom()` — queries all three traffic-light buttons via `standardWindowButton:`, converts their frames to Makepad's coordinate system via `convertRect:fromView:`, and returns the bounding box as a `Rect`.
- `win32_window.rs`: Populate `window_chrome_buttons` with the right-aligned 138×29 px bounding box of the three Makepad-drawn caption buttons.
- `linux_wayland.rs`: Populate `window_chrome_buttons` in the `WindowGeomChange` handler when `custom_window_chrome: true`, using the same right-aligned 138×29 px layout.
- `cx_api.rs`: Add `update_caption_bar_height_script_value()` to push a measured height into `mod.widgets.CAPTION_BAR_HEIGHT` on the script heap.
- `window.rs` (DSL): Declare `mod.widgets.CAPTION_BAR_HEIGHT = 27.0` in the `script_mod!` block and change `caption_bar.height` from the hardcoded `27` to `(mod.widgets.CAPTION_BAR_HEIGHT)`. On `WindowGeomChange`, derive the default caption bar height from `window_chrome_buttons` using equal top/bottom padding (`pos.y * 2 + size.y`) and trigger a script reapply.
- `platform/src/lib.rs`: Export `LinuxWindowParams`, `WindowGeom`, and `SafeAreaInsets`.


kevinaboos:
@wheregmis once this is merged in, you should be able to properly position the show/hide sidebar buttons that are in makepad-studio's caption bar.

PR #1004, opened 2026-04-01
Sync caption bar controls to window geometry

wheregmis:
Add logic to compute and apply left/right insets so caption bar controls don't overlap window chrome buttons. Introduces STUDIO_CAPTION_CONTROL_GAP and caption_bar_control_insets(window_geom) to calculate margins based on chrome button position, plus App::sync_caption_bar_controls and App::sync_main_window_caption_bar_controls to update and redraw views. Wire syncing into startup, WindowGeomChange handling in both actions and events paths. Includes unit tests covering default, macOS-style (left-side) and Windows-style (right-side) chrome layouts.

Waiting for the regression after #1001 so will rebase afterwards, test and put it for review


wheregmis:
Will do somthing more on studio in future, but closing it for now

PR #1031, opened 2026-04-10
fix(widgets): fix Toggle hover overriding active state color

TigerInYourDream:
## Problem

When a Toggle is in the ON (active) state and the user hovers over it, the pill background flashes to the hover color (gray) instead of staying in the active color (e.g., blue). This makes the toggle appear to lose its active state on hover.

## Root Cause

The Toggle's pixel shader in `widgets/src/check_box.rs` has an inconsistent mix chain order:

```glsl
// color_fill / color_stroke (BUG): active → hover
// hover comes AFTER active, so it overrides the active color
.mix(self.color_active, self.active)
.mix(self.color_hover, self.hover)     // ← overwrites active

// mark_color (CORRECT): hover → active
// active comes AFTER hover, so it correctly overrides hover
.mix(self.mark_color_hover, self.hover)
.mix(self.mark_color_active, self.active)  // ← correctly overrides hover
```

## Fix

Swap the `hover` and `active` lines in the `color_fill` and `color_stroke` mix chains to match the `mark_color` order:

```glsl
.mix(self.color_hover, self.hover)     // hover first
.mix(self.color_active, self.active)   // active overrides hover
```

This ensures:
- **OFF + hover**: shows `color_hover` (correct hover feedback)
- **ON + hover**: shows `color_active` (stays in active color)
- **ON + no hover**: shows `color_active` (unchanged)

## Scope

This PR **only** fixes the Toggle pill's `color_fill` and `color_stroke` mix chains. The following are pre-existing issues not addressed here:

- **CheckBox pixel shader** (lines 84-85, 91-92) has the same `active → hover` order — separate concern, different widget variant.
- **`draw_text.get_color`** in CheckBoxFlat (line 138-139) has the same order — affects the Toggle's internal label text, which is typically empty (`text: ""`). Left for a follow-up if needed.
- **`mark_color_active_hover` uniform** is defined but unused in the mark_color mix chain — pre-existing, unrelated to this fix.

Reported via: Project-Robius-China/robrix2#78


kevinaboos:
Is this correct? It seems that a style change _should_ occur when hovered over, even if the toggle element is enabled (active).

TigerInYourDream:
With the old order, when you hover over a toggle that's already ON, the pill visually reverts to its OFF appearance — it looks like the toggle lost its active state just because the cursor passed over it. This PR fixes that by ensuring active overrides hover in the mix chain (matching what `mark_color` already does).

A proper active+hover visual feedback (e.g. a `color_active_hover` uniform) would be a separate enhancement.

TigerInYourDream:
To clarify: the issue is the mix **order**, not the color values. When `hover` comes after `active` in the chain, `mix(..., 1.0)` fully replaces the active color regardless of what `color_hover` is set to. The only app-level workaround would be setting `color_hover` = `color_active`, which then breaks OFF+hover feedback.

kevinaboos:
Right, i agree with that, but the change you proposed here would give no visual feedback that the user is hovering over an activated/selected toggle.

imo, adding a `color_active_hover` is the only proper solution

PR #1032, opened 2026-04-10
Add table rendering to TextFlow engine for Markdown and Html widgets

poborin:
Implement streaming-compatible table support in the shared TextFlow layout engine. Tables render incrementally as events arrive — no buffering needed since pulldown_cmark provides column count upfront via the header separator line.

- Add FlowBlockType::TableCell variant with SDF shader for cell borders and header backgrounds in DrawFlowBlock base definition
- Add begin/end table, row, and cell methods to TextFlow using turtle layout with equal-width column distribution
- Draw cell borders after row layout completes so all cells share the row's max height for uniform borders
- Replace 8 TODO stubs in markdown.rs with direct TextFlow calls
- Add table/thead/tbody/tr/th/td tag handling in html.rs with column counting via HtmlNode lookahead
- Skip whitespace-only text nodes inside HTML tables using the pre-computed all_ws flag

Before:
Screenshot 2026-04-10 at 5 04 12 pm

After:
Screenshot 2026-04-10 at 9 32 11 pm

An issue with rendering strikethrough text is tracked at #1033

### DIFF
--- a/widgets/src/html.rs
+++ b/widgets/src/html.rs
@@ -137,50 +137,10 @@ script_mod! {
             quote_fg_color: theme.color_label_inner
             code_color: theme.color_bg_highlight
             selection_color: theme.color_selection_focus
+            table_header_bg_color: theme.color_bg_highlight
+            table_border_color: theme.color_shadow
             space_1: uniform(theme.space_1)
             space_2: uniform(theme.space_2)
-
-            pixel: fn() {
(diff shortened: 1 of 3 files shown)


kevinaboos:
nice, thanks for adding this, will test it in Robrix

kevinaboos:
thanks for adding this. can confirm that it works well in Robrix!

PR #1037, opened 2026-04-11
set surface_alive=true at startup

dai1975:
Recent patch introduces surface_alive state, but it seems not set at startup.

This patch simply set it, but there may be more suitable way(I don't understand whole codes).

### DIFF
--- a/platform/src/os/linux/android/android.rs
+++ b/platform/src/os/linux/android/android.rs
@@ -1740,6 +1740,7 @@ impl Cx {
 
             //cx.maybe_warn_hardware_support();
 
+            cx.os.surface_alive = true;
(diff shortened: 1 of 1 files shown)


kevinaboos:
I think my PR #1035 should cover this issue already, what do you think?

dai1975:
Wmm, latest commit 3cf602a4322f8eb4b5cf8709aeac13e6913f7713 still show black screen at startup time on android.

kevinaboos:
yep, you're right, I noticed that yesterday as well during some testing. I just did a full day of fixes on this (with a lot of other android improvements) and have just opened PR https://github.com/makepad/makepad/pull/1043. Hopefully that works for you too!

dai1975:
perfectly work. thank you.

PR #1091, opened 2026-05-21
Overhaul android tooling and platform layers to support Android 8

kevinaboos:
(includes some commits from #1089, please merge that first)

---------------

Major redesign of the Android platform and Android tooling in cargo-makepad to support many new things:
* target SDK is now API level 35, which is required for google play store submissions
* Support generating an `.aab` file (Android APp bundle or something), which is the new required format for submitting to the Play store 
* min SDK is now API level 26 (android 8, Oreo) 
* compile SDK is unchanged (still 33)
* tested robrix across 5 devices, working well)
* huge refactors to android platform layers to not unconditionally assume that certain features will be available on a given platform layer
   * thihgs like libamidi, choreographer, etc 
* major fixes to virtual keyboard/IME, edge-to-edge layout support (immersvie vs non-immersive)
   * improve IME scrolling animation to lock it to the edge of the actual virtual kbd
   * avoid double padding (related to safe inset areas) on older devices pre-API 35
* Support setting the system bar appearance to theme-based color (dark vs light mode)

### DIFF
--- a/libs/toml_parser/src/toml.rs
+++ b/libs/toml_parser/src/toml.rs
@@ -47,6 +47,10 @@ pub enum Toml {
     Num(f64, TomlSpan),
     Date(String, TomlSpan),
     Array(Vec<Toml>),
+    /// An inline table `{ k = v, ... }` encountered in a value position (e.g.
+    /// an array element). Top-level inline tables in `key = { ... }` form are
+    /// instead flattened into dotted keys by `parse_key_value`.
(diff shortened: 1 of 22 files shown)


kevinaboos:
actually i'm still making some minor improvements to the AAB tooling, nothing within Makepad platform itself.

kevinaboos:
ok done, ready to merge

PR #1101, opened 2026-06-03
Support standard keyboard nav shortcuts/keys in TextInput

kevinaboos:
Implement platform-standard TextInput navigation and deletion behavior, including Home, End, PageUp, PageDown, word movement, line/document boundaries, and Shift-based selection.

* Use Apple Option/Cmd conventions on Apple targets
* Use Ctrl conventions on non-Apple targets
* Web accepts both shortcut styles for now, since we don't have a way to query the host OS from within a makepad web env.

Also, be extremely careful to ensure that we respect Unicode grapheme boundaries when doing all the selection/navigation logic.

Fix `Delete`, which was erroneously handled before.

Add lots of missing keys in Linux X11 & Wayland backends, e.g., Home, End, Delete, Insert, PageUp/PageDown, and arrow keys


kevinaboos:
marked as draft until the upstream commit error in makepad is fixes

PR #1118, opened 2026-06-09
Codex Plan in Makepad AI

wheregmis:
I have a codex plan, so just wanted to lay foundation before doing anything on studio


wheregmis:
I verified that we dont use makepad_ai lib in studio, and majority of this is under hub. So removing it right now

PR #1119, opened 2026-06-10
Linux: fix desktop close/terminate lifecycle handling 

kevinaboos:
Please merge after #1117.

This avoids a strange case where Linux (both X11 and wayland but
in different ways) failed to properly shut down cleanly.

* Fix X11 close handling by removing destroyed windows from the map,
  which helps avoid delivering duplicate window closed events.
* Restructure how close handling happens on Wayland too, and allow
  the app to respond to a close request just like other platforms.
* Add a Linux second-signal hard exit so that repeated Ctrl+C
  or `kill` comands can actually terminate a failed/hung shutdown.
* Also a tiny fix to windows too: clear Win32 `GWLP_USERDATA` during
  `WM_DESTROY` to avoid accessing an old window pointer.

### DIFF
--- a/draw/src/turtle.rs
+++ b/draw/src/turtle.rs
@@ -1708,7 +1708,9 @@ impl<'a, 'b> Cx2d<'a, 'b> {
                     let max = max.eval_width(self);
                     turtle = self.turtles.last_mut().unwrap();
                     if let Some(max) = max {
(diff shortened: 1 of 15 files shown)


kevinaboos:
ah, this can be its own standalone PR with no other deps. Will close & resubmit.

PR #1127, opened 2026-06-17
New unified kinetic scrolling. Vastly improve draw shaprness on low-res screens. Fix rendering and text perf issues on CPU+GPU

kevinaboos:
This PR is huge, but unfortunately is rather difficult to split up into smaller PRs due to the multi-point fixes that have to happen simultaneously across multiple subsystems / platform backends. In general, the goal is to:
1. improve low-res screen appearance: help prevent blocky/pixellated things like emoji/avatars, identified in https://github.com/project-robius/robrix/issues/926
3. fix perf issues across all backends
4. re-do scrolling entirely, to make it feel native and to unify it across portallist, regular scrollable views, and all other widgets. Add a kinetic scroll model based on chrome, android, and iOS.
5. general rendering correctness on older/strict GPUs
6. text/image caching performance, particularly for HTML
7. an absolute shit ton of per-platform windowing and input fixes

Below is a complete generated summary, including breaking changes (there aren't too many) to be aware of. I've already fixed up all the in-repo examples, experiments, and makepad-studio to ensure that they all work with this PR's changeset.

---------------------------

## ⚠️ Breaking changes

There aren't too many breaking changes, just four main ones to know about. Only (1) and the new `#[live]` Splash fields are relevant to most makepad users.

**1. `PortalList::max_pull_down` is removed** (live DSL — fails at *runtime*, not compile time)

Overscroll is no longer a fixed distance cap; it's a rubber band whose stretch follows the
gesture's momentum, gated per edge. `max_pull_down` is replaced by `bounce_at_start` /
`bounce_at_end` (both `#[live(true)]`).

```diff
  list := PortalList {
-     max_pull_down: 0.0
+     bounce_at_start: false   // 'no pull-down at the top' — the exact equivalent
  }
```

Because makepad parses `live_design!` at runtime, a stale `max_pull_down` **compiles fine and
then errors on app startup** (`property max_pull_down not defined on type`). Grep your DSL.
Non-zero values have no exact equivalent — the stretch now follows the gesture and flattens at
35% of the viewport. `FlatList` keeps its own `max_pull_down`.

**2. `tessellate_path_fill` / `tessellate_path_stroke` take a trailing `tolerance: f32`**

(`makepad_draw::vector`, draw/src/vector/triangulate.rs.) Curve flattening is now sized in
device pixels rather than path units. Pass `0.25` to keep the previous behavior.

**3. `Font::glyph_outline_bounds_in_ems` takes `&mut Option<Rc<GlyphOutline>>`**

(was `&mut Option<GlyphOutline>`.) Glyph outlines are shared via `Rc` instead of deep-copied
every frame. Use the new `Font::glyph_outline_rc()`, or unwrap the `Rc`. `Font::glyph_outline()`
still returns an owned `GlyphOutline` and is unchanged.

**4. `ScrollEvent` / `FingerScrollEvent` gain a `phase: ScrollPhase` field**

Only affects code that builds these structs with a literal (synthetic events, tests, custom
backends). Use `phase: ScrollPhase::None` to preserve old behavior; `ScrollPhase` is exported
from `makepad_platform::event`.

### Silently changed behavior (compiles, parses, but acts differently)

- **`flick_scroll_scaling` and `flick_scroll_decay` are now dead.** They still exist as `#[live]`
  fields and still parse, but nothing reads them. Deceleration is tuned with the new `fling_decel`.
  Anything tuning scroll feel through the old two fields silently stops having an effect.
- **`flick_scroll_maximum` default raised `80.0` → `240.0`** and its unit changed: it is now a
  release *speed* in per-frame px at a nominal 60fps (×60 → px/s, so 14,400 px/s).
- **Both edges now bounce by default.** Lists that never had a bottom overscroll now get one.
- **`PortalListAction::Scroll` now fires for *all* viewport motion** (auto-tail, programmatic
  scrolls, draw-side renormalization), not just user input — it moved from six event-handler
  emit sites to a single draw-side emit. Code using `scrolled()` as a "the user scrolled" proxy
  should gate on user input explicitly. Opt out entirely with `emit_scroll_actions: false`.
- **`ScrollBars::get_scroll_pos()` can now return values outside `[0, max]`** while a bounce is
  animating (it carries the rubber-band overscroll). Clamp if you depend on the old range.

## New `#[live]` DSL fields

All additive — existing DSL keeps working.

| Widget | Field | Default | Purpose |
|---|---|---|---|
| PortalList, ScrollBar | `fling_decel` | `0.997` | per-ms velocity decay of a flick |
| PortalList, ScrollBar | `bounce_at_start` | `true` | rubber-band past the start edge |
| PortalList, ScrollBar | `bounce_at_end` | `true` | rubber-band past the end edge |
| PortalList | `emit_scroll_actions` | `true` | emit `PortalListAction::Scroll` on viewport moves |
| PortalList | `reached_start_margin` | `Some(0)` | one-shot start-edge sentinel |
| PortalList | `reached_end_margin` | `Some(0)` | one-shot end-edge sentinel |
| Window | `draw_ssaa_resolve` | — | SSAA downscale-resolve shader |

### Scrolling & kinetic motion (all platforms)

- **Replace the per-frame flick decay with a shared, frame-rate-independent fling model**  
  _Feature_  
  Touch/mouse-drag flicks used to sum the per-sample velocities of the last (up to) 4 samples, scale by `flick_scroll_scaling`, and then multiply a per-frame pixel delta by `flick_scroll_decay` once per frame — so the same flick travelled a different distance at 60Hz vs 120Hz, its speed depended on how many events happened to land in the 4-sample buffer, and on a 500Hz+ mouse those 4 samples spanned a few milliseconds, amplifying jitter into full-speed flings from tiny drags. The new `widgets/src/scroll_motion.rs` (shared by ScrollBar and PortalList, so every scrollable view now decelerates identically) estimates release velocity like a native VelocityTracker (100ms of retained history, capped at 64 samples, velocity taken over the last 40ms) and integrates `v *= decel^elapsed_ms` per frame, with each frame's `dt` clamped to an EMA band (0.5x-1.5x, hard cap 0.1s) so one late or early frame — e.g. Windows `Present(1,0)` returning early or spanning two vblanks — becomes a slightly uneven step instead of a visible jump or stall. Tuned values change meaning: `flick_scroll_scaling` and `flick_scroll_decay` still compile but are now ignored (tune deceleration with the new `fling_decel`), and `flick_scroll_minimum`/`flick_scroll_maximum` are now read as per-frame pixels at a nominal 60fps (x60 -> px/s), with `flick_scroll_maximum`'s default raised 80.0 -> 240.0 (= 14,400 px/s) on both ScrollBar and PortalList.

- **Hard flicks carry a native distance, and a quick re-flick boosts the fling**  
  _Feature_  
  A flat exponential decay tuned to feel good for gentle scrolling sheds a hard flick's speed several times faster than iOS or Android do, so a hard flick died almost immediately and deceleration bit the instant the finger lifted. The fling now blends its decay rate from the widget's base rate toward a slower one as release speed rises (2,000 px/s -> 8,000 px/s, 0.997 -> 0.9985) and opens fast flicks with a ~0.25s near-constant-speed glide, so a hard flick reaches its speed and glides a near-native distance while gentle scrolling keeps the base feel. A press that catches a fling now also remembers its speed: a same-direction re-flick within 0.4s adds it back (clamped to `flick_scroll_maximum`), so repeated flicks build up speed as they do in Chrome on Android. An exact port of Android's `OverScroller` fling spline is included behind a `USE_ANDROID_FLING_SPLINE` const (currently off: a maximum fling coasts ~2.6s and its high-speed tail felt too strong) for A/B testing the native curve.

- **Add rubber-band overscroll at the scroll limits, with per-edge opt-out**  
  _Feature_  
  Content now stretches past its scroll limits and springs back, using the curve that matches the input: a finger (touch or mouse drag, and the flicks they release) gets the loose iOS `UIScrollView` model — an asymptotic stretch that flattens at 35% of the viewport and a critically damped spring back — while trackpad momentum gets Chrome's stiffer macOS elastic model (`elastic_overscroll_controller_exponential`). A bounce is seeded by a drag released past the edge, a fling that reaches the edge, or a trackpad momentum stream that hits it; the seed velocity is softened against the remaining headroom so a huge flick peaks smoothly under the 35% cap instead of slamming flat against it, and a dedicated `FrameClock` advances its first frame by a nominal 1/60s so the fling-to-spring hand-off doesn't freeze one frame at fling speed. The band only engages on an axis that actually has content to scroll, so the unused direction of a two-axis view never stretches, and a plain wheel event or a force-stop (scrolling blocked, or the scrollbar handle grabbed) clears any stretch and kills the spring rather than freezing a stretch on screen.

- **Make trackpad scrolling gesture-aware so momentum goes to the view that was actually scrolled**  
  _Feature_ · _macOS, Linux/Wayland_  
  ScrollBar now switches on the scroll event's `ScrollPhase`. Finger-driven deltas (`Began`/`Changed`) reset the gesture and stop any running fling — so putting fingers back on the pad catches the scroll — and a new `owns_gesture` flag records whether this bar actually moved on the latest finger-driven delta. That fixes nested scroll views: a bar pinned at its limit does not own the gesture, so the OS `Momentum` deltas that follow are not applied by it and chain up to the ancestor instead of being silently swallowed by the inner view. Momentum that arrives once a bar has pinned at an edge is dropped rather than consumed invisibly, so a press on visibly stationary content is an ordinary click and not a fling catch, and the remaining momentum can feed the rubber band instead. macOS reports the full phase set (`Began`/`Changed`/`Ended`/`Momentum`/`MomentumEnded`/`Touched`); Wayland reports `Changed`/`Ended` only (no OS momentum stream). Phase-less wheel input (`None`, i.e. Windows, X11, Web and classic mice) keeps the existing `smoothing`/`set_scroll_target` path unchanged.

- **A press catches a running fling and no longer clicks the widget under the finger**  
  _Feature_  
  Tapping a coasting list to stop it also activated whatever button was under the finger, because the press still reached the children. `ScrollBars::catch_fling_on_press` (new) tests the raw MouseDown/TouchStart against the content rect — not `event.hits`, so it fires even where a child would capture the press — stops the fling on both axes with one press, and reports that it consumed the event; `View` then skips dispatching that press to its children (widgets/src/view.rs:783). It applies to any tap, click or touch regardless of `drag_scrolling` or finger count. On a macOS trackpad the contact and its tap-to-click arrive as separate events, so a `Touched` contact that stopped live motion arms a single-use marker (0.4s) that the following press consumes — and because the stream's end event and the touch can be delivered in either order, a stream cut within 100ms of the touch still counts as motion that touch stopped.

- **Model the OS momentum stream as an explicit state machine so parked flings survive pagination**  
  _Fix_ · _macOS_  
  The momentum lifecycle was tracked with a family of booleans and timestamps whose stale combinations repeatedly misreported whether content was actually moving, which in turn made presses either eat clicks or fail to catch a scroll. `MomentumStream` (Idle / Expected / Live / Pinned / Cut, in `scroll_motion.rs`, consumed by PortalList) makes every transition explicit and carries evidence forward — a live coast, a fresh cut — instead of letting it be erased by whichever event happens to be delivered first. `Pinned` is what makes a flick survive pagination: a stream that reaches a non-bouncing edge keeps consuming deltas with no visible effect (and no redraws), so when pagination grows the content past that edge the stream's next delta moves the list again and the user's original flick simply continues. Liveness everywhere requires a recent delta (200ms), so a stream rerouted mid-coast (pointer or window routing changed, no final event) expires instead of staying armed for a later, unrelated stream.

- **Stop swallowing every press in the first 150ms of an app's life**  
  _Fix_  
  The "a press during active scrolling settles the scroll rather than clicking through" check tracked the last finger-scroll as a plain `f64` defaulting to `0.0`, which reads as "this widget was finger-scrolled at time zero" — and since app clocks also start at zero, every press in the first 150ms of an app's life was eaten as a scroll-stop instead of reaching the widget under it. In both ScrollBar and PortalList it is now an `Option`, tested through the shared `press_settles_finger_scroll`, where `None` (nothing has ever scrolled this widget) means no press can be settling anything; negative gaps are rejected too, since a press stamped before the scroll it would be settling means the two came from different clocks and would otherwise swallow presses indefinitely. Covered by unit tests in `scroll_motion.rs`.

### PortalList

- **Rewrite PortalList's fling as a time-based, frame-rate-independent momentum model shared with ScrollBar**  
  _Feature_  
  The old flick applied a fixed per-frame delta decayed by `flick_scroll_decay` (0.97/frame), so the same flick travelled a different distance at 60Hz vs 120Hz and any dropped frame stole distance. `ScrollState::Flick` now holds a `scroll_motion::Fling` (velocity in px/s, decay per elapsed millisecond with a slower rate and an opening glide for hard flicks, jitter-smoothed frame clock), seeded from a native-style velocity tracker (`estimate_release_velocity`) over the last ~4 finger samples instead of the old ad-hoc `scaled_delta * flick_scroll_scaling`. The model lives in the shared `scroll_motion` module, so PortalList and ScrollBar (and thus ScrollBars / ScrollX/Y views) decelerate identically. `flick_scroll_maximum` is reinterpreted as a velocity ceiling (per-frame-at-60fps x60 -> px/s) and its default is raised 80.0 -> 240.0 (14,400 px/s) so hard flicks actually carry; `flick_scroll_minimum` (0.2, ~12 px/s) keeps its default but is now the release/stop speed threshold. `flick_scroll_scaling` and `flick_scroll_decay` are retained as unread `#[live]` fields purely so existing DSL keeps compiling.

- **Replace the fixed `max_pull_down` cap with a real rubber band that bounces at both edges**  
  _Feature_  
  Overscroll used to be a hard clamp: `first_scroll` was pinned to `max_pull_down` px at the top, released with a crude `*= 0.85` per-frame decay, and the bottom edge did not overscroll at all. The pull is now a proper rubber band -- the stretch is damped against the viewport extent (looser iOS curve for a finger, stiffer linear one for a trackpad; capped at 0.35 x viewport via RUBBER_BAND_TOUCH_RANGE), a drag moving back toward the content unwinds along the same curve so the content tracks the finger, and release hands the remaining velocity to a spring (`rubber_band_bounce`, seeded via `soften_bounce_velocity` so huge flicks compress into the available headroom rather than flying off). The bottom edge now bounces too, via a new `bounce_overshoot` the draw applies as an extra upward shift. `max_pull_down` is gone, replaced by `bounce_at_start` / `bounce_at_end` toggles; a non-bouncing edge pins flush, and wheel/no-phase scrolling never bounces.

- **Honor native trackpad scroll phases: OS momentum is applied as delivered instead of being re-simulated**  
  _Feature_ · _macOS, Linux/Wayland_  
  `Hit::FingerScroll` now switches on `ScrollPhase`. Finger-driven deltas (`Began`/`Changed`) apply directly and stretch the rubber band; on `Ended` any stretch springs back and an OS momentum stream becomes `Expected`; `Momentum` deltas are applied exactly as the OS sends them, so the coast has the platform's native feel and deceleration rather than makepad's own approximation; `Touched` (a finger landing on the pad) stops kinetic scrolling instantly, matching the native catch. A `MomentumStream` state machine (`Expected`/`Live`/`Pinned`/`Cut`/`Idle`) tracks the stream so a stray or already-caught stream can't restart motion, and a coast that reaches a bouncing edge hands its remaining velocity to the rubber-band spring instead of letting the stream (which can outlive the edge by a second or more) keep pushing. In practice the momentum/`Touched` half only engages on macOS, which is the sole backend that emits those phases today; Wayland delivers `Changed`/`Ended`, so it gets the finger-driven stretch and spring-back. Phase-less input (mouse wheel on Windows/X11/Web) takes the old direct-delta path and clips at the edges.

- **A press on a moving list stops it instead of clicking through; a quick re-flick boosts the fling**  
  _Fix_  
  Previously a tap during a trackpad coast (where `scroll_state` is `Stopped` because the OS owns the motion) fell through to the item under the finger, so catching a fast-scrolling list activated a button. Presses are now checked against `motion_live()` -- any unparked fling, bounce, smooth-scroll, tail animation, live OS momentum stream, or finger-driven scroll within the last 0.15s -- and are consumed as a stop, with child fan-out suppressed for the whole gesture. A trackpad tap that stops the coast has its `Touched` phase and its later press correlated (`touch_caught_motion_at`, 0.4s `CATCH_PRESS_WINDOW`) so the lift isn't delivered as a click either, and `suppress_child_events` is now also cleared on the raw `MouseUp` as a backstop, since the `FingerUp` hit that used to be its only reset depends on capture state and can be missed (leaving the flag stuck and eating the *next* click). The caught fling's velocity is remembered briefly, so a same-direction re-flick within 0.4s (`FLING_BOOST_MAX_DWELL`) adds it back and repeated flicks build up speed like native scrollers. The pulldown bounce is deliberately left running through a press -- stopping it would freeze the list with an open gap at the edge.

- **Flings and trackpad momentum survive pagination instead of being killed at the edge**  
  _Feature_  
  When a flick or coast hit a non-bouncing edge it used to be terminated. For an infinite list that loads history on reaching the top, that meant the flick died exactly when the load was triggered and the newly prepended items arrived to a stationary list. The fling is now *parked* (and the OS stream *pinned*) at such an edge: it keeps decaying silently -- no movement, no redraws, presses are ordinary clicks -- and if content grows past the edge the very next step moves the list again, so the original flick continues naturally through the load.

- **Emit Scroll once per frame from the draw, and add one-shot ReachedStart/ReachedEnd edge actions**  
  _API_  
  `PortalListAction::Scroll` was fired ad hoc from each event handler (wheel, flick step, pulldown step, tail step, smooth-scroll step, scroll-bar), so it could fire several times a frame, and never fired for draw-side renormalization. It is now emitted from a single end-of-draw comparison of `(first_id, first_scroll, bounce_overshoot)` -- at most once per frame, for every kind of motion -- and can be switched off entirely with `emit_scroll_actions` on lists that don't consume it. New `ReachedStart`/`ReachedEnd` actions fire once when an end of the list comes on screen (re-armed when it leaves), so infinite lists can paginate on a sentinel instead of watching scroll positions; the margins let them prefetch N items early. `PortalListRef::reached_start()`/`reached_end()` are the accessors. `scrolled()` and `smooth_scroll_reached()` now scan all of the widget's actions rather than only the first one -- with several action kinds emitted per frame, `find_widget_action` could return a different action and wrongly report `false`. `PortalListRef::set_first_id_and_scroll()` now routes through the inner method, which clamps away impossible overscroll at a non-bouncing start edge and re-arms the edge announcements (as does `set_first_id`).

- **Stop auto-tail and scroll-bar feedback from fighting an active fling, drag, or bounce**  
  _Fix_  
  Both the tail auto-scroll (which re-asserts position by subtracting `overflow` from `first_scroll` every draw) and the scroll bar (which re-emits a position quantized to the nearest item/offset via the height tree) were writing `first_scroll` back underneath a running animation, snapping it back by up to ~1px per frame. At the slow end of a flick, where the real per-frame motion is sub-pixel, that snap-back dominated and the content visibly shook in place instead of gliding to a stop. Both are now skipped while a gesture or animation owns the scroll position, and any pending tail adjustment is discarded rather than applied on the first post-fling frame. Symmetrically, explicit navigation now takes clean ownership: keyboard nav (Home/End/PageUp/PageDown/Arrow Up/Down), a scroll-bar drag, and `smooth_scroll_to()` all call `stop_all_scroll_motion()` first, so no leftover fling, bounce, or OS momentum stream resumes from wherever the viewport lands.

- **Cut per-frame work in height tracking and mouse-move fan-out**  
  _Perf_  
  `HeightTree::update` now returns whether the height actually changed (or was measured for the first time), and lists backed by a height tree fold the item height into the running average only then -- previously every visible item re-recorded its height on every draw, which both cost work per frame and skewed the average toward whatever happened to be on screen longest. `update_default_height` re-applies the default to every unmeasured item, an O(unmeasured) walk, and now only runs once the average has drifted by 0.5px rather than 0.001px. Separately, a `MouseMove` outside the list's clipped rect is no longer fanned out to the whole item subtree: one rect test guards it, with the first outside-move still delivered so the last hovered item gets its hover-out.

- **Fix bottom-aligned short lists snapping their content to the top on alternating draws**  
  _Fix_  
  With `align_top_when_empty: false` and content shorter than the viewport, the draw deliberately leaves a gap above the first item -- but the edge-pin logic treated any positive `first_scroll` as overscroll and clamped it to 0 whenever the scroll state was `Stopped`, so the content jumped between bottom-aligned and top-aligned on successive draws. The pin now computes that resting gap explicitly and leaves it alone, clamping only genuine overscroll.

### Widgets: View, Dock, Splitter, Window, Label, Icon

- **Window: stop re-resolving caption-bar geometry on every mouse move (fixes scroll judder on Windows)**  
  _Perf_  
  `Event::WindowDragQuery` fires once per `WM_NCHITTEST` — i.e. on every mouse move on Windows — and each one did two widget-tree lookups (`caption_bar`, `windows_buttons`) plus area→rect resolution, which showed up as jitter when the mouse moved during a fling. The resolved `(visible, caption_rect, buttons_rect)` is now cached and only recomputed on relayout: invalidated on `WindowGeomChange` and whenever the caption-layout inputs (`show_caption_bar`, `caption_bar_height_override`, system caption height) change, e.g. a live/DSL reload. The cache is only populated once the caption bar and its visible buttons have actually been laid out, so an early query can't pin a zero `buttons_rect` and make the min/max/close strip respond as draggable caption (a click on Close would have dragged the window).

- **Window: skip caption redraws when the title and centering padding are unchanged**  
  _Perf_  
  `sync_caption_title` runs on every event via `ensure_initialized` and unconditionally cloned the title and pushed it through `set_text`, costing a widget-tree label lookup per event. It now remembers the last synced title and bails out early (also handling the case where the caption label doesn't exist in the tree yet, retrying on a later event); `Event::LiveEdit` clears the memo since a live reload can re-apply DSL text over the label. `sync_caption_centering` likewise only writes the label's left padding and redraws when the padding actually moved.

- **Label: `set_text` is a no-op for identical text; new `set_text_color` setter**  
  _API_  
  `Label::set_text` previously rebuilt the string and called `redraw` even when the text was unchanged, dirtying the draw list on every call — callers that re-push the same string each event (e.g. the window caption title) now cost nothing. Adds `Label::set_text_color(cx, Vec4f)` and `LabelRef::set_text_color`, which also short-circuit when the color is unchanged.

- **View: a press that catches an in-progress fling is fully consumed and no longer clicks through**  
  _Fix_  
  Tapping a scrolling view to stop momentum also activated whatever child widget was under the finger (a button, a list row). `View::handle_event` now asks the scroll bars `catch_fling_on_press()` before dispatching to children; when a fling was caught, the press is neither forwarded to children nor run through this View's own hit handling (`FingerDown`/animator), matching iOS, Android and macOS behavior. The flag is only ever set on a press (MouseDown / touch Start) over a flinging scroll view, so key/hover handling is untouched. Backed by the new `widgets::scroll_motion` module.

- **Splitter/View: dragging a splitter now actually resizes cached (draw-list-optimized) panes**  
  _Fix_  
  Two halves of one bug. `View::walk_from_previous_size` pinned every non-`Fill` size to the previous frame's measurement, so a `Fixed` size that the parent had just re-resolved (exactly what a splitter drag produces) compared equal in the cached-draw dirty check and the pane never re-laid-out; `Fixed` sizes are now passed through live alongside `Fill`, and only content-driven sizes fall back to the previous measurement. On top of that, `Splitter` redrew only `self.a`/`self.b`, which are empty in Dock usage (the dock draws the pane contents itself), so it now also calls `cx.redraw_area_and_children` on both pane areas to reach nested cached views.

- **Dock: closing the active tab left the closed tab's pixels on screen**  
  _Fix_  
  When the active tab was closed, the next tab usually lands at the same selected index, so `select_tab`'s index-unchanged shortcut skipped the redraw even though the displayed item had changed. `close_tab` now explicitly redraws the tab bar's `contents_draw_list` after re-selecting.

- **Window: opt-in full-window supersampling (SSAA) via `MAKEPAD_SUPERSAMPLE`**  
  _Feature_  
  New `SsaaStack` renders the entire UI — including the overlay (tooltips, modals, context menus) — into an offscreen pass at `dpi * N` and downscales it with a single fullscreen resolve quad, where one bilinear tap is a 2x2 box average at N=2. This sharpens icons, emoji and images on low-DPI screens. Only the render-target pixel density and viewport are inflated; the pass rect stays the logical window size, so layout and hit-testing are unchanged. It is off by default (analytic AA already covers most of it and SSAA is costly) — set `MAKEPAD_SUPERSAMPLE=2` to enable, clamped to <=4 and read once into a `OnceLock` because `begin`/`end` query it per window per frame. Skipped when gauss capture is active for the window, to avoid nesting the two scene mechanisms.

- **Icon: stop clipping icons, which squared off round stroke caps**  
  _Fix_  
  The `Icon` DSL now overrides the inherited layout defaults with `clip_x: false` / `clip_y: false` (both default to `true`). SVGs are fit to their content bounds, which do not include stroke extent, so the tight clip cut the outer half of round caps off. Visual change for anyone who relied on an Icon clipping its content to its walk.

- **Window: the caption bar scales with UI zoom on platforms whose buttons we draw ourselves**  
  _Fix_ · _Windows, Linux/Wayland_  
  The caption bar height was pinned to the platform-reported window-chrome-button geometry on every platform that reports it (macOS, Windows, Linux/Wayland). On Windows and Linux/Wayland those buttons are drawn by makepad and scale with the UI zoom, so the bar stayed at native height while its buttons grew or shrank. The height is now derived from the system geometry only on macOS, where the buttons are OS traffic lights that don't zoom and the title has to line up with them; elsewhere the caption bar's `height: Fit` lets it grow along with the zoomed buttons.

### Image & texture caching

- **Mipmap decoded images so minified avatars and thumbnails stop aliasing into a blocky mess**  
  _Feature_  
  `ImageBuffer::into_new_texture` now uploads non-animated decoded images as `VecMipBGRAu8_32` (mipmapped) instead of single-level `VecBGRAu8_32`, so a large source image minified into a small on-screen rect (room avatars, inline thumbnails) samples a mip chain instead of being bilinearly undersampled and aliasing badly. Animations are deliberately excluded because frames must not bleed across mip levels. Where the backend uploads CPU-built levels, the chain comes from an alpha-weighted 2x2 box filter (RGB premultiplied by alpha so fully-transparent texels can't bleed undefined color into the average). Gated by a new public `makepad_platform::texture::image_cache_use_mipmaps()`: default on for OpenGL-on-Linux builds (`all(target_os = "linux", not(use_vulkan))`) only, and forced on or off on any backend via the `MAKEPAD_IMAGE_MIPMAPS` env var (`1`/`true`/`on`/`yes` enable, anything else disables).

- **Only build a CPU mip chain on backends that actually upload one, and build it on the decode worker**  
  _Perf_  
  The box-filtered chain grows decoded-image memory by about a third (4/3x) and costs a full multi-pass filter per image, so it is now only constructed when it will be read: `backend_uploads_cpu_mip_chain()` is true for Metal (macOS/iOS/tvOS) only. OpenGL generates the chain on the GPU from level 0 via `glGenerateMipmap` and just needs the level count, so `mip_chain_max_level()` computes `max_level` analytically without allocating any levels; the remaining backends (Vulkan, D3D11, WebGL) upload level 0 only. Where the chain is built, `build_cpu_mip_chain_if_uploaded()` runs inside `spawn_decode_job` on the decode thread pool, keeping the filter cost off the UI thread instead of paying it when the texture is committed (it stays idempotent, so the sync/headless path still gets a chain). A unit test pins `mip_chain_max_level` to the level count `generate_bgra_mip_chain` actually emits.

- **Image: stop showing the previous occupant's picture in recycled widgets while a new source decodes**  
  _Fix_  
  In a `PortalList` (or any widget reuse), an `Image` recycled for a new source kept displaying the old texture, and on `dev` the "still loading" draw branch bound *no* texture at all, so the shader kept whatever was last left in `draw_vars` — a ghost image from an unrelated item. `Image` now records texture provenance in a new `#[rust] texture_async_source: Option<PathBuf>` and routes all three async entry points (`load_image_file_by_path_async`, `load_image_from_data_async`, `load_image_http_by_url_async`) through one `begin_async_load` / `finish_async_load` / `cancel_async_load` state machine. A texture whose provenance is a *different* async key is cleared when a new load starts, while content the widget legitimately owns — a `set_texture` placeholder such as a blurhash, or a previous load of the same key — deliberately stays visible during the decode. The loading draw path now explicitly binds either that texture or `empty_texture(0)`, never stale draw vars. Related fixes in the same state machine: a completed decode is only applied if the widget is still waiting on that exact key (`async_image_path` is now cleared once consumed instead of lingering); a failed load blanks the widget rather than leaving someone else's image up; `set_texture` and `load_svg_from_data` cancel any pending raster load so a late decode can't overwrite the new content; and `ImageRef::set_texture` routes through the `ImageCacheImpl::set_texture` impl instead of assigning `inner.texture` directly, so these invariants hold for every caller.

- **Image: stop the endless redraw loop when a widget re-issues the same async load every frame**  
  _Fix_  
  Widgets that call `load_image_*_async` on every draw (list items repopulated per frame) hit the image cache, got `AsyncLoadResult::Loaded` back, and unconditionally called `redraw(cx)` — dirtying themselves into a redraw loop that never settled. `finish_async_load` now early-returns when there is no pending load and the requested path is already the source of the currently bound texture, so re-loading the image that is already displayed skips both the redraw and the `async_load.off` animator and is a genuine no-op.

- **Accept any `Arc<D: AsRef<[u8]> + ?Sized>` as async image data instead of only `Arc<Vec<u8>>`**  
  _API_  
  The async data path is now generic over `Arc<D> where D: AsRef<[u8]> + Send + Sync + ?Sized + 'static`: the free `image_cache::load_image_from_data_async`, the `ImageCacheImpl::load_image_from_data_async_impl` default method, the internal `spawn_decode_job`, `Image::load_image_from_data_async`, and `ImageRef::load_image_from_data_async`. Callers that already hold image bytes as `Arc<[u8]>` — a common shape when the bytes are shared with a network or storage layer — no longer have to copy them into a fresh `Arc<Vec<u8>>` just to hand them to the cache. Existing `Arc<Vec<u8>>` call sites still compile unchanged (`D` infers to `Vec<u8>`), and the trait method carries `Self: Sized` so `ImageCacheImpl` remains object-safe.

- **Add `Image::has_content()` / `ImageRef::has_content()`**  
  _API_  
  `has_texture()` only reports a raster texture, so an `Image` displaying an SVG answered `false` and callers using it to decide whether to draw a fallback got it wrong. `has_content()` returns true for either a raster texture or a loaded SVG, distinguishing "the load was accepted" from "there is actually something to draw". A pending async load does not make it false: content deliberately kept visible during the decode (a `set_texture` placeholder, or a previous load of the same key) still counts.

### HTML & TextFlow widgets

- **Html: stop rebinding link/span sub-widgets to the wrong HTML nodes**  
  _Fix_  
  Custom sub-widgets (and `<details>` blocks) inside `Html` were keyed by a visit-order counter incremented during the draw walk. Because a collapsed `<details>` skips its hidden nodes, toggling one renumbered every widget after it, so links and spans silently re-bound to a different node's attributes (wrong href/text). They are now keyed by `node.index`, the node's stable index in the parsed doc, which does not shift with skip mode. Two related leaks are closed: `TextFlow::item_with_scope` now recreates the widget when the template stored at an entry id differs from the requested one (it previously returned the stale widget, so an `<a>` replaced by a `<span>` at the same position kept rendering as a link — `item_with` already did this), and `Html::set_text` now calls `text_flow.clear_items()` when the parsed doc actually changes (mirroring `on_after_apply`), so a recycled `Html` widget (e.g. in a PortalList) no longer keeps sub-widgets that captured their attributes from the *previous* document.

- **TextFlow: don't panic on unbalanced HTML close tags**  
  _Fix_  
  `end_code()` and `end_quote()` did `area_stack.pop().unwrap()`, so a stray `</pre>` or `</blockquote>` with no matching open tag — trivially reachable from untrusted input such as a chat message — panicked the app. Both now return early when the stack is empty instead of ending a block that never began. Relatedly, `begin_list_item()` pushed onto `area_stack` while `end_list_item()` never popped, leaking one entry per list item and letting a stray close tag pop a list item's area into the code/quote block; that push is removed, so `area_stack` is now balanced and pushed/popped solely by begin/end_code and begin/end_quote.

- **HtmlLink: browser-standard link colors, and make pressed_color reachable**  
  _Fix_  
  The hover branch was tested before the pressed branch, but the animator's `hover.pressed` state also sets `hovered: snap(1.0)`, so `pressed_color` could never win. Precedence is now pressed > hovered > base, and a state color that is `None` falls back to the base `color` rather than pushing nothing — previously an `HtmlLink` with no `hover_color` dropped to the ambient TextFlow font color on hover, i.e. the link stopped looking like a link exactly when you pointed at it. The DSL defaults now come from two new exported consts, `HTML_LINK_COLOR` (#0000EE) and `HTML_LINK_PRESSED_COLOR` (#EE0000), and the default `hover_color` (#00EE00) is removed to match real browsers. Caveat: `HtmlLink`'s animator `hover.on` state applies `pressed: snap(1.0)` as well as `hovered: snap(1.0)` (unlike `Button`, whose `on` state zeroes `down`), and this change does not touch that animator — so with the stock animator a merely-hovered link now satisfies the pressed branch and renders `pressed_color`, and `hover_color` stays unreachable until that state is corrected.

- **Html/TextFlow: skip the re-parse when set_text is given identical HTML**  
  _Perf_  
  Callers commonly re-populate widgets with the same content every draw frame (list item recycling, streaming updates). `Html::set_text` now returns immediately if the body string is unchanged, avoiding a full `parse_html` per frame; when the string does change, it only tears down state (`clear_items`, `seen_details`, `summary_area_cache`) if the newly parsed doc actually differs from the old one. Both paths mean a re-populate that yields the same document preserves sub-widgets and user-toggled `<details>` open state instead of collapsing back to the initial state.

- **Add runtime color setters for TextFlow and HtmlLink**  
  _API_  
  New `TextFlow::set_font_color()`, `set_code_color()` and `set_quote_bg_color()` taking `Vec4f` (the latter two write `draw_block.code_color` / `draw_block.quote_bg_color`, covering inline code, code blocks and quote backgrounds), plus `HtmlLink::set_color()` / `HtmlLinkRef::set_color()` taking `Option<Vec4f>` (`None` = inherit the surrounding text color). All are no-ops when the color is unchanged, so they are safe to call unconditionally from a theme-update path. Because an `HtmlLink` is drawn by its parent `TextFlow` rather than owning a single area, `set_color` redraws each area in the link's `drawn_areas` so multi-line links repaint fully.

### Text: layout, shaping, rasterization, atlases

- **Cache the layout of long texts instead of re-laying them out on every frame**  
  _Perf_  
  The layout cache refused to admit any text longer than 512 bytes, or longer than 192 bytes with 4+ lines (`Layouter::should_cache_text`), so long chat messages and code blocks were re-shaped and re-laid out from scratch every single frame they were on screen. That admission filter is gone; the cache is now a recency-refreshing LRU bounded by a new 16 MB soft byte budget (`LAYOUT_CACHE_MAX_BYTES`, weighted by text bytes plus laid-out row/glyph storage) on top of the existing 4096-entry cap. Entries drawn in the current frame are never evicted, so one frame's visible set cannot evict itself into a permanent every-frame-miss cycle, and the over-budget excess is reclaimed at the frame boundary (`Layouter::advance_cache_generation`, called once per frame from `Fonts::prepare_textures`) one frame after the text stops being drawn rather than lingering until the next insert. Four unit tests cover admission of long texts, byte-budget eviction, working-set protection, and frame-boundary reclaim.

- **Make the shape cache a true LRU so hot words survive scrolling through varied text**  
  _Perf_  
  The shaper's 4096-entry cache evicted in pure insertion order (a `VecDeque` FIFO) and cache hits did not refresh recency, so words that were being re-shaped every frame got dropped purely because they were old, forcing repeated rustybuzz shaping while scrolling. Hits now bump the entry to the front of an LRU order keyed by a monotonic tick.

- **Stop deep-copying (and rescanning) glyph outlines for every drawn glyph**  
  _Perf_  
  `Font` now caches `Rc<GlyphOutline>` and exposes `glyph_outline_rc()`. The rasterizer's SDF and MSDF paths previously cloned the outline's full command `Vec` out of the cache for every glyph on every draw — even when the atlas slot already existed and no rasterization happened — and the MSDF path additionally rescanned the whole command list per call to estimate outline complexity. Outlines are now shared by `Rc` and the complexity metric is computed once when the outline is built (`GlyphOutline::complexity()` / `OutlineComplexity`, moved out of `rasterizer.rs`); the only remaining deep copy is the owned outline handed to the MSDF worker thread, which happens at most once per glyph per atlas epoch.

- **Downsample color-emoji strikes toward display size so emoji stop looking blocky on low-DPI screens**  
  _Quality_  
  Color emoji ship as a single large PNG strike, which was decoded into the color atlas at native size and then minified by the GPU with no mipmaps — visibly blocky at normal text sizes, worst on 1x displays. `GlyphRasterImage::decode_scaled` now box-downsamples the strike at decode time (alpha-weighted, so fully transparent texels don't bleed their undefined color into the average) to ~1.5x the display size, never upscaling, and `origin_in_dpxs`/`dpxs_per_em` are rescaled by the achieved integer-rounded scale (geometric mean of the per-axis ratios) so on-screen size and position are unchanged; the atlas slot shrinks to match, freeing color-atlas space. Target sizes are quantized into 1.25x buckets rounded up toward native, so a continuous zoom or a fractional-DPI change reuses an existing atlas slot instead of re-decoding the full strike and allocating a fresh slot at every sub-pixel size step.

- **Upload only the newly appended rows of the SLUG glyph atlas when it grows**  
  _Perf_ · _Windows, Linux/Wayland, Linux/X11_  
  The append-only SLUG atlas textures marked themselves fully dirty on any dimension change, so every frame that appended a glyph re-uploaded the entire RGBA-f32 texture (up to ~16 MB) — a significant contributor to laggy/juddery scrolling. Only a width change now forces a `TextureUpdated::Full`; a height growth leaves all existing rows in place and emits a `Partial` dirty rect covering just the appended rows. Both the D3D11 and the desktop-Linux GL backends honor this: each keeps spare height capacity (~3x the needed height, with a floor) for the atlas and uploads only the new rows, via `UpdateSubresource` and `glTexSubImage2D` respectively, instead of recreating the texture on every append. In-place reuse is restricted to pure appends past the upload frontier so an in-flight frame's curve data can't be overwritten. Android/OHOS GLES keep the conservative full `glTexImage2D` re-upload (the float-atlas sub-upload path renders emoji as black boxes there), and Metal re-uploads the whole texture regardless, so behavior on those is unchanged.

### Vector, SVG & shader rendering quality

- **Size SVG anti-aliasing and curve flattening in device pixels instead of path units, so icons stay sharp at any viewBox/DPI**  
  _Fix_  
  DrawSvg now derives a device_scale (fit-scale x dpi; geometric mean of the two axis scales when preserve_aspect is off) and hands the tessellator an AA fringe of 1.0/device_scale (clamped to [0.2, max(content/6, 4.0)]) and a flatten tolerance of 0.05/device_scale (clamped to [0.01, 0.25]), via new DrawVector #[rust] fields cur_fill_aa / cur_stroke_aa / cur_tolerance (0.0 keeps the old defaults: fill aa 1.0, the caller's stroke aa, tolerance 0.25); svg/render.rs now prefers cur_stroke_aa over the old `stroke_width.min(1.0)` heuristic. Previously the fringe and tolerance were constants in path-local units, so the same icon was fat/blurry at one viewBox and a hard aliased step at another (a 512-unit viewBox drawn at 24px got a sub-pixel fringe), and because the stroke tessellator builds half-width as `w/2 + aa/2` (libs/svg/src/tessellate.rs), an over-wide fringe fattened every stroke (robrix #926). Static SVGs still tessellate once and reuse the uploaded cache; they now re-tessellate only when device_scale drifts more than 5% (new #[rust] DrawSvg::cached_scale), e.g. on a DPI change or resize.

- **Draw round line caps as solid discs; they no longer look square at small stroke widths**  
  _Fix_  
  In libs/svg/src/tessellate.rs the round-cap triangle fan emitted its rim vertices with u = u0 (which is 0.0 whenever aa > 0), and the vector fragment shader maps stroke coverage as `sd = 1 - abs(u*2 - 1)`, so u = 0 means zero coverage: the semicircle was a radial fade from the solid fan center out to nothing, and at small sizes it vanished entirely, so round-capped strokes rendered like butt/square caps (robrix #926). Both emit_cap_start and emit_cap_end now emit the arc points at u = 0.5 (full coverage), matching the fan center.

- **Fix the seam across the middle of Sdf2d box_y shapes with different top/bottom corner radii**  
  _Fix_  
  box_y's distance field used only `length(max(q,0)) - 2r`, which is a constant -2r everywhere in the interior, so when the mix switched from the top radius to the bottom radius at the vertical midpoint the field jumped discontinuously and left a visible horizontal coverage seam wherever interior distance is actually consumed (blur, stroke_keep's `abs(shape) - width`, glow). It now adds the standard `min(max(q.x,q.y), 0.)` interior term for both radii, making the field continuous through the midpoint. (box_x still has the same shape of field and is unchanged.)

- **Prefix Metal/WGSL shader locals with `l_` so user variable names can't collide with reserved words**  
  _Fix_ · _macOS, iOS, Linux/Wayland, Linux/X11_  
  ShaderBackend::map_local_name emitted non-shadowed locals verbatim for the Metal and WGSL targets (the HLSL and GLSL arms already prefixed them; shadowed names already got an `_s<n>` prefix), so a shader local named `half`, `kernel`, `constant`, `thread`, or any of WGSL's larger reserved set produced source that failed to compile at runtime. Declarations, references and (via map_param_name, which delegates for these targets) function params all route through map_local_name, so the rename is internal to codegen and invisible to DSL authors. Metal covers macOS/iOS; the WGSL backend feeds the Linux Vulkan/naga (use_vulkan) shader path.

### Platform core (draw lists, events, uniforms)

- **Add gesture phases to scroll events so widgets can tell user-driven trackpad deltas from OS momentum**  
  _API_  
  `ScrollEvent` and `FingerScrollEvent` gain a `phase: ScrollPhase` field (plumbed through `Event::hits`), with `ScrollPhase` being a new enum: `None` (default), `Began`, `Touched`, `Changed`, `Ended`, `Momentum`, `MomentumEnded`. This is what lets the new scroll/fling model in the widgets crate follow the OS deceleration curve on a trackpad instead of guessing it, and stop a coasting fling the moment a finger lands on the pad. macOS populates the full set, including the macOS-only `Touched` (a raw finger contact with a zero delta, emitted for every touch even one that never becomes a scroll gesture, so it must not disturb tap-to-click or drag, which arrive separately as mouse events). Wayland reports `Changed` per pointer frame and `Ended` on finger lift-off for touchpads, and `None` for wheels. X11, Windows, Web, headless and the studio stdin transport always report `None`, so widgets that ignore the phase keep their existing behavior.

- **Fix `DrawCallUniforms`/`DrawListUniforms::as_slice()` returning arrays 4x too long (out-of-bounds uniform reads)**  
  _Fix_  
  Both `as_slice()` methods declared their return type as `&[f32; size_of::<Self>()]` and got there by `transmute` on the reference -- but `size_of` is in bytes, not f32 elements, so the returned array was four times the real struct: `[f32; 16]` for the 16-byte `DrawCallUniforms` and `[f32; 96]` for the 96-byte `DrawListUniforms`. Reference transmutes don't check the pointee size, so this compiled silently, and every backend that uploads these (Metal, D3D11, OpenGL, Vulkan, WebGL, headless raster) read three quarters of its uniform payload past the end of the struct (undefined behavior) and uploaded those bytes into the GPU constant buffers. The lengths are now `size_of::<Self>() >> 2`.

- **`set_zbias()` now reports whether the depth actually changed, so cached draw calls stop being rejected by the depth test**  
  _Fix_  
  The zbias is the fragment depth and is handed out from the global draw order on every frame, so it shifts whenever any draw list drawn before a given draw call grows or shrinks -- including for a cached draw call whose draw list was not redrawn and therefore has `uniforms_dirty` unset. A backend that uploads `DrawCallUniforms` only on `uniforms_dirty` then leaves a stale depth on the GPU and the LESS_EQUAL depth test drops that draw call, making cached content disappear. `set_zbias` now returns `bool` instead of `()`; D3D11 (the one backend that uploads conditionally, and where the bug was visible) uploads when it returns true or when the constant buffer was never uploaded at all. The other backends upload draw-call uniforms unconditionally and ignore the result, so they are unaffected; existing callers that ignore the return value still compile.

- **GPU pass metrics size mipmapped texture uploads from the whole mip chain, not just level 0**  
  _Quality_  
  `estimate_texture_upload_bytes`, which feeds `GpuPassMetrics::texture_bytes` from `Cx::collect_gpu_pass_metrics` (currently consumed by the Android backend's per-frame metrics), computed `width * height * bpp` for `VecMipBGRAu8_32` and `VecMipRGBAf32`, counting only level 0 rather than the full mip chain (~4/3 of level 0) that actually gets uploaded; it now derives the size from the actual data buffer length when the buffer is present and falls back to the level-0 estimate when it isn't. Diagnostics only, no rendering change. Separately, `makepad_platform`'s root re-export list now exposes `image_cache_use_mipmaps()` (which the draw crate picks up via `pub use makepad_platform::*`) so the image-cache path can consult the mipmap policy: default on for Linux/OpenGL only, overridable with `MAKEPAD_IMAGE_MIPMAPS`.

### Windows (D3D11 / Win32)

- **Windows: pace presentation to the display refresh (vsync) with a DXGI frame-latency waitable swap chain**  
  _Perf_ · _Windows_  
  The D3D11 backend always presented uncapped (`Present(0, ..)`) from a free-spinning `EventFlow::Poll` loop, so during any scroll or animation it rendered far more frames than the monitor could show (e.g. ~127 fps on a 99 Hz panel). Since makepad's scroll/fling animations advance a fixed step *per rendered frame*, DWM discarding the surplus frames unevenly made scrolling visibly judder — the "laggy scroll on Windows" report. Main windows are now created with `DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT` + per-swap-chain `IDXGISwapChain2::SetMaximumFrameLatency(1)`, the render loop waits on the waitable (33 ms timeout so an occluded window can't stall) before building each frame, and `handle_repaint` presents with vsync by default; setting `MAKEPAD_NO_VSYNC` (any value) opts out, mirroring the Linux/EGL backend. The old device-level `IDXGIDevice1::SetMaximumFrameLatency(1)` is removed so the two mechanisms can't conflict; popups keep latency 1 but no waitable.

- **Windows: live-resize and swap-chain lifetime fixes for the new waitable swap chain**  
  _Fix_ · _Windows_  
  Resize frames present with sync interval 0 and rely on `DwmFlush()` for compositor sync (a blocking `Present(1)` would add up to a full refresh interval of latency to every resize frame, making live resize feel heavy), and they skip the frame-latency wait. Because retired presents still credit the waitable semaphore, `stop_resize()` now drains the accumulated credits — otherwise the per-frame wait would return immediately forever after the first resize and pacing would be silently dead for the rest of the window's life. `ResizeBuffers` now passes the same creation flags the chain was made with (a mismatch fails with `DXGI_ERROR_INVALID_CALL`), and `D3d11Window` gained a `Drop` impl that `CloseHandle`s the waitable, which is an app-owned kernel handle and was otherwise leaked per window lifecycle.

- **Windows: event loop paints every pass with a bounded message-drain budget instead of only when the queue empties**  
  _Fix_ · _Windows_  
  The Poll branch peeked exactly one message and only painted when the queue was empty. Animation frame-advance (`call_next_frame_event`) and the vsync `Present` both live in the Paint callback, and a moving mouse injects `WM_MOUSEMOVE` at 500-1000 Hz, so the queue almost never emptied during a fling: the fling step was starved until the mouse paused, then jumped — the "scroll freezes/judders while moving the mouse during deceleration" bug. The loop now drains messages up to a budget (32 messages or 2 ms) and then always paints once per pass; it stays paced by the vsync Present and skips the paint only if a dispatched message requested Exit (avoiding a double-shutdown callback).

- **Windows: coalesce consecutive WM_MOUSEMOVE and WM_MOUSEWHEEL messages**  
  _Fix_ · _Windows_  
  High-polling-rate mice flooded the pump with moves, each one running a full hover hit-test over the widget tree; free-spinning wheels and precision touchpads emitted wheel messages faster than the now vsync-paced loop consumed them, so a gesture built a backlog that kept replaying deltas after it ended. Adjacent moves collapse to the newest, and adjacent wheels merge into one message carrying the summed (saturating) delta with the newest position/key-state — only *adjacent* messages are merged (peek-and-stop at the first non-matching message), so nothing is dropped or reordered past a button/key event. macOS gets move coalescing from Cocoa for free; Android already coalesced touch-moves. Note this discards intermediate cursor positions within a run, so a widget needing every pointer sample (freehand ink) would have to read raw input. Windows wheel messages carry no gesture phase, so scroll events from them are tagged `ScrollPhase::None`.

- **Windows: stop repainting at ~125 Hz while idle; sleep in GetMessageW instead**  
  _Perf_ · _Windows_  
  The event callback unconditionally returned `EventFlow::Poll`, and the 8 ms signal-poll tick recursively fired a full `Paint` — so an idle app repainted at ~125 Hz and dominated CPU. The callback now returns `Poll` only while there is real work (dirty pass, pending redraw, queued NextFrame) and `Wait` otherwise, so an idle app blocks in `GetMessageW` at ~0% CPU. The signal tick drains `platform_ops` directly (keeping Ctrl+C / SIGTERM termination, now with an `Event::Shutdown` dispatch) and still polls game controllers — a gamepad generates no Win32 message, so nothing else would wake the loop on a button press. A Paint that presents nothing but has a NextFrame armed (e.g. a video player polling between decoded frames) sleeps 1 ms, since without a Present there is no blocking call to pace the loop.

- **Windows: cache the DPI factor and the WM_NCHITTEST drag-query result**  
  _Perf_ · _Windows_  
  `get_dpi_factor()` syscalled `GetDeviceCaps` every call, and it is hit several times per `WM_NCHITTEST` and per `WM_MOUSEMOVE`; `WM_NCHITTEST` is OS-sent on every mouse move (uncoalesced) and each one dispatched a full `WindowDragQuery` event through the widget tree, often several times for the same cursor position per frame. The DPI is now cached and invalidated on `WM_SIZE`/`WM_DPICHANGED`, and the drag-query answer is memoized per raw lparam (cursor screen position), invalidated on window geometry change and on `WM_MOVE` (which the code previously ignored, leaving a stale screen-space cache when the window moved under a stationary cursor). A generation counter guards against a reentrant invalidation during the `WindowDragQuery` dispatch being clobbered by the stale result.

- **Windows: stop blocking the UI thread on game-controller enumeration every poll**  
  _Perf_ · _Windows_  
  `XInputGetState` on a *disconnected* slot is pathologically slow (it stalls on device/USB enumeration), and `poll()` probed all 4 slots on every event-loop callback — tens of milliseconds of UI-thread block, a dominant cause of scroll judder. Connected slots are still polled every call; empty slots get a one-shot full scan on the first poll (so a controller plugged in at launch is found immediately) and then one round-robin re-probe per 512 polls (~16 s worst case at the idle 125 Hz poll rate to detect a hot-plug). DirectInput `EnumDevices` (also tens of ms) is now scheduled explicitly at first poll plus every 8192 polls, phase-offset (`% 8192 == 256`) so it never lands on the same poll as an XInput probe; its `enum_timer` previously only advanced inside its own branch, so the intended "every 200 polls" throttle never actually recurred — the counter now advances once per poll.

- **Windows: update Vec textures in place via UpdateSubresource instead of recreating them**  
  _Perf_ · _Windows_  
  Every texture change destroyed and recreated the whole D3D11 texture + SRV (up to ~16 MB for the glyph atlases), a per-change hitch while scrolling; GL (`glTexSubImage2D`) and Metal (`replaceRegion`) already updated in place. D3D11 now honors `TextureUpdated::Partial` with a `D3D11_BOX` sub-rect upload, and the append-only SDF glyph atlas is allocated with 3x height headroom (min 512, rounded up to a multiple of 128) so growth reuses the texture instead of reallocating during long flick-scrolls. In-place reuse of the SDF atlas (`VecRGBAf32`) is deliberately restricted to pure appends: its shader reads per-glyph curve *counts* from the texels, so overwriting rows an in-flight frame is still sampling could feed garbage counts and hang the GPU (multi-second TDR). Adds `ID3D11DeviceContext::UpdateSubresource` to the vendored windows-rs bindings.

- **Windows: fix R8G8 texture uploads using half the correct row pitch**  
  _Fix_ · _Windows_  
  `TextureFormat::VecRGu8` (`DXGI_FORMAT_R8G8_UNORM`) was uploaded with bytes-per-pixel 1 instead of 2, so the row pitch was half the real row stride and the texture came out sheared/garbled. Also, null/zero-sized Vec texture data now returns early instead of `unwrap()`-panicking.

- **Windows: fix cached draw lists vanishing due to stale zbias uniforms**  
  _Fix_ · _Windows_  
  The zbias (fragment depth, handed out from global draw order) was advanced *after* the early-`continue`s in `render_view`, so the sequence depended on which draw calls happened to be dirty that frame rather than on the draw tree. Worse, the draw-call uniform buffer was only uploaded when `uniforms_dirty` was set, but a cached draw call's zbias still shifts whenever an earlier draw list grows or shrinks — leaving the GPU with a stale depth that the LESS_EQUAL depth test then rejects, i.e. content silently disappearing. The zbias advance is now hoisted above the early-outs and the upload condition is `uniforms_dirty || zbias_changed || buffer.is_none()` (`set_zbias` now returns whether the value changed; only D3D11 consumes it).

- **Windows: budget D3D11 shader-object creation to 4 per frame**  
  _Perf_ · _Windows_  
  Creating many D3D11 shader objects in one frame — e.g. when scrolling brings a lot of new content types into view at once — caused a visible tens-of-ms hitch. `hlsl_compile_shaders` now creates at most 4 shader objects per call across both paths combined (async-compile results and cache-hit sync creation share one budget); the remainder is queued (a `ready_backlog` deque for async, re-inserted into `compile_set` for sync) and a redraw is requested so the rest materialize over the following frames. Widgets whose shader isn't ready yet already skip their draw via the existing `os_shader_id.is_none()` guard and appear a frame or two later.

- **Windows: fix panic on Ctrl+V when the clipboard holds non-text content**  
  _Fix_ · _Windows_  
  The paste path `unwrap()`ed `GetClipboardData(CF_UNICODETEXT)`, which fails whenever the clipboard holds an image or files — pasting after copying an image crashed the app. It now skips the paste on failure, null-checks `GlobalLock` before copying, closes the clipboard on every path, and ignores the `GlobalUnlock` result (it reports "failure" with `NO_ERROR` on the normal last unlock, which the old code `unwrap()`ed).

- **Windows: don't panic on mipmapped image textures**  
  _Fix_ · _Windows_  
  `TextureFormat::VecMipBGRAu8_32` (emitted when `MAKEPAD_IMAGE_MIPMAPS` is enabled; off by default on Windows) hit the catch-all `panic!()` in the D3D11 Vec-texture upload. It now uploads level 0 as a single-level texture — safe, but no sharpness benefit yet; real per-level mip upload on D3D11 remains a TODO (Metal has it).

### Linux (OpenGL / Wayland / X11)

- **GL: stop using sampler objects and drive filtering from the texture object, fixing blocky minified textures on Mesa**  
  _Fix_ · _Linux/Wayland, Linux/X11, Android_  
  `opengl_create_samplers` no longer creates any GL sampler objects (sampler 0 is bound instead), so min/mag filter and wrap now come from the texture object state set in `update_vec_texture`/`update_render_target` — some Mesa drivers silently ignore a sampler object's `MIN_FILTER` and point-sample minified textures, which is why emoji, avatars and downscaled images looked blocky. Render-target textures (sampled by the Gaussian blur and `CachedView`) also had to move from `NEAREST` to `LINEAR` min/mag, since the sampler object used to supply that filter. Consequence: per-shader sampler declarations (`SamplerFilter`/`SamplerAddress`) are no longer honored by the GL backend — textures use LINEAR + CLAMP_TO_EDGE (or NEAREST where the texture format demands it), which matches the `ShaderSampler` default, so a shader asking for Repeat/MirroredRepeat would now clamp.

- **GL: fall back to non-mipmapped filtering when glGenerateMipmap fails, and log driver info once**  
  _Fix_ · _Linux/Wayland, Linux/X11, Android_  
  On stricter GLES3 drivers, `glGenerateMipmap` on the unsized BGRA internal format used by `VecMipBGRAu8_32` image textures raises `INVALID_OPERATION` and generates nothing; the resulting mipmap-incomplete texture samples as opaque black, so mipmapped images/avatars rendered as black boxes. The path now drains stale GL errors, checks `glGetError` after generation, and on failure warns and resets `MIN_FILTER` to `LINEAR` with `MAX_LEVEL` 0. A one-time log (on the first shader program compile) of GL vendor/renderer/version/GLSL version and sampler-object availability was added, with new `gl_sys::VERSION`/`SHADING_LANGUAGE_VERSION` constants, so driver-specific reports are actionable.

- **Linux GL: grow the SLUG glyph atlas in place instead of re-uploading the whole texture on every new glyph**  
  _Perf_ · _Linux/Wayland, Linux/X11_  
  The append-only `VecRGBAf32` glyph atlas was excluded from partial texture updates, so each time scrolling pulled new glyphs in, the entire (up to ~16 MB) float atlas was recreated with `glTexImage2D` and re-uploaded — a per-change scroll hitch. It is now allocated with ~3x height headroom (min 512 rows, rounded up to a 128-row boundary) and appended with `glTexSubImage2D`, tracking allocated capacity (`gl_cap_width`/`gl_cap_height`/`gl_cap_internal_format`) and an upload frontier (`gl_uploaded_height`) so only pure appends are written in place; a width/format change, exhausted capacity, or a non-append rect still falls back to a full realloc + re-upload. This mirrors the D3D11 `UpdateSubresource` path, and is gated off for Android/OHOS GLES where float-atlas `glTexSubImage2D` is unreliable (emoji rendered as black boxes).

- **Wayland: pace frames with wl_surface frame callbacks so redrawing a hidden window can't hang the whole app**  
  _Fix_ · _Linux/Wayland_  
  Mesa implements EGL swap interval 1 by blocking inside `eglSwapBuffers` until the previous frame's `wl_surface::frame` callback fires, and compositors withhold those callbacks for occluded/minimized windows — so a continuously-redrawing window that got hidden wedged the entire event loop. Wayland now requests swap interval 0 (X11 keeps vsync at 1, unchanged) and caps the render loop itself: each present requests a frame callback before the swap, a window with a callback still in flight is skipped and left dirty until the callback arrives, `Cx::draw_pass_to_window` returns whether the swap actually committed (a failed swap sends no commit, so it must not gate future presents), and pending callbacks are cleared when a window closes. Supporting fixes: after a Paint the loop now returns `EventFlow::Wait` rather than poll-spinning while a callback is outstanding, and the Wayland event queue is explicitly flushed before the `select()` block (a `WouldBlock` treated as transient) so queued requests actually reach the compositor. `MAKEPAD_NO_VSYNC` disables pacing for benchmarking.

- **Linux: wheel scrolling now moves a fixed 60 px per detent on X11 and Wayland**  
  _Fix_ · _Linux/Wayland, Linux/X11_  
  Both backends replaced a timing-based acceleration guess (`1200 * (0.2 - 2*dt)`, which swung between roughly 12 px and 240 px per event purely on how events happened to batch) with a shared `PIXELS_PER_WHEEL_DETENT = 60.0` logical px — about three lines, matching common Windows/GTK defaults. Wayland now reads real detent counts by binding `wl_seat` at version 9 for `AxisValue120` (high-resolution wheels, 120 units per detent, fractional detents allowed), with `AxisDiscrete` as the pre-v8 fallback and the raw accumulated axis value if a compositor sends neither; touchpad pixel deltas pass through unchanged. Raising the seat version also required mapping the xkb keymap fd `MAP_PRIVATE` instead of `MAP_SHARED`, as `wl_seat` v7+ mandates.

- **Wayland: report touchpad scroll phases so widgets can fling on finger lift-off**  
  _Feature_ · _Linux/Wayland, Linux/X11_  
  Scroll events now carry a `ScrollPhase`: wheel scrolls report `None`, finger/touchpad scrolling reports `Changed` per pointer frame and `Ended` when `wl_pointer::AxisStop` arrives. The `Ended` event is dispatched even when its delta is zero, which is what lets the scroll widgets start their own momentum animation — Wayland compositors, unlike macOS, do not synthesize momentum scrolling for clients. X11 core-protocol wheel buttons carry no gesture info and report `ScrollPhase::None`.

- **Linux: coalesce mouse-move floods on X11 and Wayland so a moving mouse no longer judders a fling**  
  _Perf_ · _Linux/Wayland, Linux/X11_  
  A 500–1000 Hz mouse queues one motion event per poll report, and dispatching each ran a redundant `WindowDragQuery` plus hover hit-test through the entire widget tree, stealing frame budget and visibly juddering an in-progress scroll fling. X11 now peeks ahead and merges only *adjacent* `MotionNotify` events for the same window, stopping at the first non-motion so no button/key/scroll event is ever dropped or reordered; Wayland buffers the latest motion and flushes it once after the dispatch batch drains — and ahead of any Button, Leave or Scroll, so hover position and event ordering stay correct — discarding it if the window was closed mid-batch. This mirrors the Windows `coalesce_mouse_move`; intermediate cursor positions are dropped, which is correct for hover/hit-testing but loses the full pointer path.

- **Wayland: UI zoom survives maximize/tile instead of resetting or flickering**  
  _Fix_ · _Linux/Wayland_  
  The Wayland-side window geom was being overwritten with the zoom-converted (layout) geom, so the next Configure read the zoomed `dpi_factor` back as if it were native — the zoom drifted or reset and the window flickered on maximize/tile. The Wayland window now keeps its native geom (it drives the buffer/viewport size and the next resize's dpi) while only the `Cx` window receives the zoomed geom, and the size/scale-changed redraw check is taken in native units before conversion.

### macOS & iOS

- **macOS: report trackpad scroll gesture phases so widgets can follow native momentum instead of guessing**  
  _Feature_ · _macOS_  
  `MacosWindow::send_scroll` now derives a `ScrollPhase` from `NSEvent.phase`/`NSEvent.momentumPhase` and puts it on every macOS `ScrollEvent` (a new additive `phase` field, defaulting to `ScrollPhase::None`): Began/Changed/Ended for the finger-driven part, Momentum/MomentumEnded for the OS momentum stream after lift-off, and None for classic wheel mice (which report 0 for both). This lets scroll widgets track real gesture boundaries — start a fling exactly on finger lift-off and decelerate at the OS rate while momentum deltas keep arriving — rather than inferring it from delta timing. NSEventPhaseCancelled (the system stole the gesture, e.g. a system-wide swipe) deliberately maps to `Began`, which widgets treat as a gesture reset, so an aborted swipe does not launch a fling the way `Ended` would.

- **macOS: touching the trackpad now stops a coasting scroll, and a tap during momentum still clicks**  
  _Feature_ · _macOS_  
  The window's view now opts into raw indirect trackpad touches (`setAllowedTouchTypes: NSTouchTypeMaskIndirect`) and implements `touchesBegan/Ended/CancelledWithEvent:`. Every raw finger contact emits a zero-delta `ScrollEvent` with the new `ScrollPhase::Touched`, which widgets use to halt kinetic scrolling the instant a finger lands — previously a coast could only be caught by an actual scroll or click, and neither is reliable, since macOS suppresses tap-to-click while its momentum stream is live and that stream can end while a widget's own deceleration tail is still gliding. To close the same hole for clicking, `MacosWindow` synthesizes the tap macOS swallowed: a touch that starts within 0.25s of the last momentum event, stays single-finger, lasts under 0.3s, and sees neither a scroll gesture nor a real mouse-down is delivered as a PRIMARY down/up pair. Outside the momentum window the touch is disqualified up front so the OS-delivered click is never doubled.

- **iOS: stop losing rotations and safe-area changes that UIKit delivers re-entrantly, and never deliver Init twice**  
  _Fix_ · _iOS_  
  UIKit fires resize/safe-area notifications synchronously, sometimes while the app state is already borrowed or an event callback is on the stack; such a notification used to be silently dropped, leaving the app laid out for the old size — potentially forever, because the display link is paused while idle so nothing forces a redraw, and the geometry diff gate would see no change on a later check. Those cases now set a `GEOM_CHECK_PENDING` flag which is drained by re-reading the view's real geometry at the tail of the next completed event dispatch (the repeating 8ms timer guarantees one runs soon); `IosApp::do_callback` returns whether the callback actually ran, so `apply_new_window_geom` can restore `last_window_geom` and re-arm the flag when its `WindowGeomChange` was dropped. A `GEOM_DRAIN_ACTIVE` guard keeps each re-check a single bounded step rather than a recursion. Separately, geometry is now only applied from `draw_in_rect` before the first frame — applying it from `drawableSizeWillChange` or `viewSafeAreaInsetsDidChange` ran the first-draw path and delivered `Init`, and then the first draw delivered `Init` again (double app startup), since only `draw_in_rect` clears `first_draw`.

- **Metal: upload mip chains for `VecMipBGRAu8_32` textures instead of uploading nothing**  
  _Fix_ · _macOS, iOS_  
  The Metal texture descriptor already sized itself for the full mip chain, but the update path had no arm for `VecMipBGRAu8_32` and fell through to `_ => 0`, so a texture in that format got allocated and never received any pixel data — not even level 0. A new `update_mip_data_bgra` walks the concatenated chain that `draw/src/image_cache.rs::generate_bgra_mip_chain` produces (level 0 first, one BGRA u32 per texel) and `replaceRegion:mipmapLevel:` uploads each level, halving the dimensions as it goes and bailing out if the buffer is shorter than the declared chain. This is what makes the new mipmapped image path usable on Apple; note the format is off by default there (image mipmaps default on only for OpenGL, override with `MAKEPAD_IMAGE_MIPMAPS`), and an in-code TODO flags that a mip-filter sampler is still needed on Metal for the levels to actually improve minification quality.

### Web / WASM

- **Web: tag browser scroll events with the new ScrollPhase as None, keeping wheel input on the classic (non-kinetic) scroll path**  
  _Quality_ · _Web_  
  The kinetic-scroll rework adds a `phase: ScrollPhase` field to `ScrollEvent` (platform/src/event/finger.rs, variants None/Began/Touched/Changed/Ended/Momentum/MomentumEnded). The web backend's `ToWasmScroll -> ScrollEvent` conversion in platform/src/os/web/to_wasm.rs fills it with `ScrollPhase::None`, since a DOM wheel event carries no trackpad gesture phase. Web scrolling is unchanged: ScrollBar and PortalList match on `phase` and treat `None` as classic wheel input, so they never enter the gesture-end fling or OS-momentum-follow paths that macOS (and, for gesture end, Wayland) now use. X11, Windows and headless send `None` for the same reason.

- **Web: comment the VecMipBGRAu8_32 texture path as level-0-only, noting real mipmaps are still a TODO on WASM**  
  _Quality_ · _Web_  
  The low-DPI sharpness work adds an opt-in CPU mip chain for decoded images (`image_cache_use_mipmaps()` in platform/src/texture.rs; default on only for Linux/OpenGL, overridable with `MAKEPAD_IMAGE_MIPMAPS`). platform/src/os/web/web_gl.rs already routed `VecMipBGRAu8_32` through the single-level BGRA upload; this branch only adds a comment marking that as the intended fallback and flagging real mips (`gl.generateMipmap`) as a TODO. Comment-only, no behavior change -- and in practice the mip format never reaches web, since the env override cannot be set under wasm.


kevinaboos:
ok, finally ready to go after a month!

PR #1165, opened 2026-08-07
macos: fix UI freezing after the window has been fully hidden

kevinaboos:
The present-gated frame pacing from commit `c0b851a57`
introduce a bug here, which basically causes vsync or other paint
events to never reach to UI after the macOS window is re-shown.
(moved to the foreground after having been bkgd'd for a while)

Fixes for this:
* Skip presenting  while the window's `occlusionState` lacks
  `NSWindowOcclusionStateVisible`, keeping the pass dirty so the first
  visible beat repaints. This also avoids `nextDrawable` hard-blocking
  the main thread (we set `allowsNextDrawableTimeout: NO`) on a pool
  exhausted by presents the compositor will never consume.
* Add a 250ms watchdog to force the event handlers to start again
  if they locked up somehow.
* On a watchdog timer, nudge `drawableSize` to rebuild the drawable pool
  the same way a manual resize does, reclaiming stuck drawables so the
  recovery present can't block on an exhausted pool.


kevinaboos:
Closing in favor of #1166

PR #1205, opened 2026-08-31
fix(script): fold `&&`/`||` used in shader argument position

Antonio-Macia:
A logic operator (`&&` / `||`) used in *argument* position inside a
shader — e.g. `max(x, float(a && b))` — pushed a `ShaderMe::LogicOp`
marker that the consuming opcode popped before `handle_logic_phi` could
fold the operands. `pop_to_me` had no arm for `ShaderMe::LogicOp`, so it
fell through to `_ => todo!()` and panicked deep inside makepad with
`not yet implemented` and no shader-source location — the DSL author got
no hint about which construct caused it.

Intercept a `LogicOp` on top of `mes` in `pop_to_me` and fold
`(first op second)` right there, mirroring exactly what
`handle_logic_phi` does at statement level, then deliver the combined
value to the enclosing expression. The fold only needs shared `vm`
access (the second operand's `Id` type is resolved via the shader scope,
like the call/builtin arms), so no signatures change. Nested logic ops
(`a && b || c`) are handled by the trailing recursive `pop_to_me`.

The generated string is byte-identical to the statement-level path, so
`&&`/`||` in argument position now compiles and renders just like it
does in a `let`. Remaining type errors (e.g. `float(bool)`) are reported
as normal located shader diagnostics, not panics.

### DIFF
--- a/platform/script/src/shader.rs
+++ b/platform/script/src/shader.rs
@@ -1634,6 +1634,51 @@ impl ShaderFnCompiler {
             self.skip_next_pop_to_me = false;
             return;
         }
+        // A logic operator (`&&` / `||`) sitting on top of `mes` here means it was used
+        // in *argument* position, e.g. `max(x, float(a && b))`. The short-circuit RHS
+        // ends right at this point, so the consuming opcode reaches us before
+        // `handle_logic_phi` got a chance to fold the operands. Fold `(first op second)`
+        // now — exactly as `handle_logic_phi` does at statement level — and deliver the
+        // combined value to the enclosing expression. Previously this fell through to a
+        // `todo!()` that panicked inside makepad with no shader-level diagnostic.
+        if matches!(self.mes.last(), Some(ShaderMe::LogicOp { .. })) {
+            if let Some(ShaderMe::LogicOp {
+                op,
+                first_operand,
+                first_type,
+                ..
+            }) = self.mes.pop()
+            {
+                // Pop the second operand (the already-evaluated RHS). Resolve an `Id`
+                // to its pod type the same way the call/builtin arms below do — using
+                // the shader scope, so we only need shared `vm` access (no `pop_resolved`,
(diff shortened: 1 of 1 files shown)


Antonio-Macia:
I went back to re-verify this against today's `dev` (`0e19a65ea`) before asking
anyone to spend time on it, and I think you should know what I found: **the
original repro no longer panics.**

The case this PR was written for was a logic operator in argument position of a
builtin cast:

```rust
let big = max(self.col.x, float(w > 10.0 && h > 10.0))
```

On June's `dev` that hit `_ => todo!()` in `pop_to_me` and panicked with `not yet
implemented`, pointing inside makepad with no shader-source location. On today's
`dev` the same shader gives a proper, located diagnostic instead:

```
[E] examples/bugfix/src/main.rs:31:65 - scalar pod constructor expects number type, got Bool
[E] draw shader 'DrawQuad' failed to compile and will NOT be drawn
```

Which is the outcome this PR was trying to produce, reached by a different route:
the pod constructor now rejects `Bool` before codegen gets far enough to reach
the unhandled `LogicOp`. I also checked the type-valid form (a logic op passed
straight to a script fn taking the flag) and that compiles fine on `dev` today,
with the folded `&&` present in the generated source.

So I can no longer demonstrate a construct that reaches the `todo!()` this patch
guards, and I would rather say so than have you merge a fold that may now be
unreachable. Two options and I am happy with either:

- **Close it.** The user-visible symptom is gone, and an unreachable arm is
  maintenance for nobody's benefit.
- **Keep it as defence in depth.** `pop_to_me` still ends in `_ => todo!()`, so
  whatever construct does reach it next will panic without a location again. If
  you would rather have that arm return a located diagnostic than a `todo!()`,
  that is a smaller and more honest change than this one, and I will rewrite it
  that way.

Your call — I have no attachment to the patch, only to the failure mode.

PR #1206, opened 2026-08-31
fix(widgets): validate show_bg + texture_caching at apply, not with a draw panic

Antonio-Macia:
A `View` with both `show_bg: true` and `texture_caching: true` (or `optimize:
Texture` directly) hit `panic!("dont use show_bg and texture caching at the same
time")` in `draw_walk` on the first frame. The two options are statically
incompatible — known the moment the live design is applied — so panicking at draw
time is late and crashes the whole app over a config mistake.

Detect the conflict in `on_after_apply`, right after `optimize` is resolved, emit
an actionable `error!` (located at the apply site, like other widget config
errors) and degrade gracefully: keep `show_bg`, drop the texture cache
(`optimize = None`). The live design still loads and the app keeps running; the
existing `draw_walk` panic is no longer reachable from a live design.

### DIFF
--- a/widgets/src/view.rs
+++ b/widgets/src/view.rs
@@ -251,6 +251,20 @@ impl ScriptHook for View {
             self.optimize = ViewOptimize::DrawList;
         }
 
+        // A view cannot both paint its own background (`show_bg`) and be cached into a
+        // texture (`texture_caching`, or `optimize: Texture` directly): the two are
+        // mutually exclusive and would otherwise hit a hard `panic!` deep inside
+        // `draw_walk` on the first frame. Catch the conflict here, at apply/load time,
+        // with an actionable error and degrade gracefully — keep `show_bg`, drop the
+        // texture cache — so the live design still loads and the app keeps running.
+        if self.show_bg && self.optimize.is_texture() {
+            error!(
+                "View: `show_bg` and `texture_caching` are mutually exclusive; ignoring \
+                 `texture_caching` and keeping `show_bg`. Remove one of the two."
+            );
+            self.optimize = ViewOptimize::None;
+        }
+
         if self.optimize.needs_draw_list() && self.draw_list.is_none() {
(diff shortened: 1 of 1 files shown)


Antonio-Macia:
Here is a runnable repro, since the PR text alone asks you to take my word for it.

Still reproduces on today's `dev` (`0e19a65ea`), Linux/OpenGL, llvmpipe:

```
thread 'main' panicked at widgets/src/view.rs:1079:25:
dont use show_bg and texture caching at the same time
```

Drop this into `examples/bugfix/src/main.rs` and run it:

```rust
pub use makepad_widgets;
use makepad_widgets::*;

app_main!(App);

script_mod! {
    use mod.prelude.widgets.*

    startup() do #(App::script_component(vm)){
        ui: Root{
            main_window := Window{
                window.inner_size: vec2(200, 200)
                pass.clear_color: #x1A1A1A
                body +: {
                    bad := View{
                        width: Fill
                        height: Fill
                        show_bg: true
                        texture_caching: true
                        draw_bg +: { color: #3B82F6 }
                    }
                }
            }
        }
    }
}

#[derive(Script, ScriptHook)]
pub struct App {
    #[live]
    ui: WidgetRef,
}

impl MatchEvent for App {}

impl AppMain for App {
    fn script_mod(vm: &mut ScriptVm) -> ScriptValue {
        crate::makepad_widgets::script_mod(vm);
        self::script_mod(vm)
    }
    fn handle_event(&mut self, cx: &mut Cx, event: &Event) {
        self.match_event(cx, event);
        self.ui.handle_event(cx, event, &mut Scope::empty());
    }
}
```

What makes this worth catching at apply time rather than at draw time: the two
properties are set in separate places in a live design, both are individually
valid, and nothing reads wrong when you write them. The failure arrives on the
first draw, as a panic inside makepad, naming a file the author never opened.

PR #1207, opened 2026-08-31
fix(draw): bound the image cache by bytes, not by entry count

Antonio-Macia:
`ImageCache` capped itself at 512 *entries*. That is the right unit for an IDE
full of 32px icons — 512 of those are about 2 MB — and the wrong one everywhere
else. The same 512 entries are ~5.6 GB of 1080p photos, or ~65 GB of 24 Mpx RAW.

The failure mode is quiet: the cap never fires, and the process grows until the
OS kills it. A limit in the wrong unit doesn't report an error; it reports
nothing. The existing comment even said so out loud — "the cap is generous
enough that this rarely triggers in practice" — which is true for icons and
exactly backwards for photographs.

A realistic case: an image grid with 240 photos stays comfortably under the
entry cap while holding ~2.5 GB.

This adds a byte budget (256 MB) and keeps an entry ceiling as a backstop for
the pathological many-tiny-images case. Eviction now takes the biggest textures
first: when the problem is memory, dropping a hundred icons doesn't help if
what's over budget is one 96 MB RAW. The old eviction removed entries in
unspecified order, which was fine when the only goal was bounding a HashMap.

Texture size counts the mipmap chain (the 4/3 series), since static images are
mipmapped on upload — counting only the base level undershoots by a third.

`insert_loaded` keeps its signature; eviction moved to a free function because
measuring a texture needs `&mut Cx` and reaching the cache goes through
`cx.get_global`, which already borrows it.

Eviction remains safe for the same reason it always was, now stated in the doc
comment: every widget holds its own clone of the texture it is displaying, so
evicting a cache entry never blanks a visible image.

### DIFF
--- a/draw/src/image_cache.rs
+++ b/draw/src/image_cache.rs
@@ -723,9 +723,33 @@ impl Default for ImageCache {
 }
 
 impl ImageCache {
-    /// Max distinct cached images before eviction kicks in. Each `Loaded` entry can hold a
-    /// decoded GPU texture, so without a cap this `HashMap` grows for the process lifetime.
-    const MAX_ENTRIES: usize = 512;
+    /// Byte budget for cached GPU textures before eviction kicks in.
+    ///
+    /// This used to be a cap on the *number* of entries (512), which is the right
+    /// unit for an IDE full of 32px icons — 512 of those are about 2 MB — and the
+    /// wrong one everywhere else. The same 512 entries are **~5.6 GB** of 1080p
+    /// photos, or **~65 GB** of 24 Mpx RAW: the cap never fires and the process
+    /// grows until the OS kills it. A limit in the wrong unit doesn't report an
+    /// error; it reports nothing.
+    ///
+    /// 256 MB is a deliberately modest default: it comfortably holds an icon set
+    /// or a page of thumbnails, and forces recycling on a photo grid — which is
+    /// exactly what you want there.
+    pub const MAX_BYTES: usize = 256 * 1024 * 1024;
+
+    /// Hard ceiling on entry count, kept as a backstop for the pathological case
+    /// (thousands of tiny images), where the byte budget alone would still let the
+    /// `HashMap` grow unbounded.
+    pub const MAX_ENTRIES: usize = 4096;
+
+    /// Bytes a decoded RGBA texture of `width * height` occupies, mipmaps included.
+    ///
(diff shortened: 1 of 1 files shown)


kevinaboos:
256MB seems pretty low, both Rik and I have valid usage scenarios where the app is showing loads of images, especially Rik.

i agree that maybe it shouldn't be capped by a number of entries, but this cap seems way too conservative. 

@makepaddev what do you think here?

Antonio-Macia:
Fair, and I think you're right. The constant is the weakest part of this PR, and it's weak for exactly the reason the old cap was: 256 MB is a number picked from *my* case — a photo grid that should be recycling — the same way 512 entries was picked from an IDE full of icons. A fixed number in the right unit is still a fixed number.

What I'd like to keep is the unit and the eviction order. The budget itself should belong to the app, not to me.

Concretely:

- Make the budget a field on `ImageCache` with a setter on `Cx`, instead of an associated const, so Robrix or anything else can raise it at startup without forking `makepad-draw`.
- Raise the default to whatever you and Rik consider sane — 1 GB, or derived from total system RAM (say 1/8, clamped), which avoids picking a number for machines none of us can see. If you'd rather not pull in a platform query for that, a plain higher default is fine by me.
- Keep the entry ceiling as the backstop for the many-tiny-images case.

Worth noting that a high default costs nothing when you're under it: eviction only walks the map once the budget is exceeded, so apps that legitimately hold thousands of images just never pay for it.

That still fixes the actual bug — a limit that never fires, so the process grows until the OS kills it, with no error anywhere — while leaving the ceiling where the app can see it.

Happy to push that change; just tell me what default you want.

kevinaboos:
Instead of spamming lengthy AI replies, can you actually respond with human words to encourage us that you've understood what this PR is doing?

Antonio-Macia:
Fair enough, that was too much text.

Short version: the cache counted entries, not bytes. 512 icons and 512 photos look identical to it, but the second is gigabytes, so the cap never fires and the process just grows. I hit it with a photo grid, not in theory.

You're right that 256MB is my number for my case. Make it settable and pick whatever default suits Robrix, and I'll write it.

And yes, to be straight with you: I don't speak English, so I use an AI to write it for me. The patch and the problem are mine, the wording was not, and it clearly came out sounding like a machine. I'll keep it short from now on.

kevinaboos:
@makepaddev please close this spam.

PR #1234, opened 2026-09-13
script, widgets: Octoscript VM hardening — bounded equality, uncaught-error termination, canonical try/catch, heap and string caps, execution caps, parser diagnostics; host mods in Splash isolates

ymote:
Octoscript (https://github.com/OctoSense-org/Octoscript) is a capability-first standalone runtime for the makepad script language: it embeds `makepad-script` to run untrusted scripts with an explicit budget for time, memory and host access. Running that way found a set of VM holes that apply to any makepad embedding that evaluates script it did not write. This series ports the fixes onto `work` (d476e5245) as seven commits; each message records the PORT / ADAPT / RETIRE decision taken against upstream's newer code.

What the series adds:
- Equality (`==`/`!=`, `ScriptHeap::deep_eq`) is an iterative worklist in `equality.rs`: cycles and shared DAGs terminate, NaN stays unequal to itself, each comparison is charged VM fuel (one unit per pair/edge/typed-array element, ceiling 65,536) and samples the hard deadline every 256 units; exhaustion is an uncatchable bail.
- An uncaught script error with no active `try` frame terminates the eval (`ScriptTrapOn::Bail`) instead of draining the diagnostic and running on. Host opt-out: evals with `silence_errors` (Splash's streaming `eval_with_append_source`) keep drain-and-continue.
- Canonical `try protected catch fallback` (legacy `try a b [ok c]` still parses), cross-call unwinding to the nearest try frame, loop back-edges discard iteration-local tries/operands/mes. Precedence fixes: `a || b && c`, comparisons bind tighter than equality, `!` negates truth instead of bit-flipping the f64.
- Heap and per-string caps expressed over upstream's `ScriptAllocationBudget` (`set_max_heap_bytes`, `set_max_string_bytes`, `reconcile_heap_bytes`), plus `ScriptStringBuffer` / `new_bounded_string_with` / `temp_bounded_string_with` for hosts that build strings incrementally; array index validation before storage is touched; field update keeps insertion order; `cast_to_f64` yields a traced NaN for non-numeric text.
- Operand-stack and call-frame caps (`with_stack_value_limit`, `with_call_frame_limit`), `ScriptNative::clear_type_methods` for restricted embeddings.
- Re-entrancy and handle hardening: `run_core` no longer holds a raw opcode pointer across native calls (a re-entrant `eval` that reloads the active body used to dangle), `ScriptThreads` validates indices in release builds, handle downcasts compare `Any::type_id`, value previews truncate on char boundaries, `libs/regex` drops a `never_loop` allow.
- Four fuzz-found parser/tokenizer panics fixed: field-assignment reverse-pair walk stopping at a 1-opcode chunk, prototype-field `:` rewrite keeping opcode/source-map lockstep and rejecting an unavailable rust-value index, `_` at a numeric boundary, and multibyte identifiers underflowing `pos`.
- Structured parser diagnostics: `ScriptParserDiagnostic`, `diagnostics()`, `diagnostics_truncated()`, `set_emit_errors()`; the `parse_errors` sink stays in lockstep.
- `register_splash_isolate_mod(fn(&mut ScriptVm))`: host code can install its own mods into every Splash isolate, run last so the ambient-authority strip cannot remove them.

Behavioural changes a makepad app author would notice:
- A plain `eval` now stops at the first uncaught error and returns it; wrap in `try … catch …` to recover. Streaming (`silence_errors`) evals are unchanged.
- `~` (LOG) and `ScriptVm::log` are refusable by the host: `ScriptVmBase::allow_debug_output` (default `true`, so nothing changes unless a host clears it); when off, `~` raises a catchable not-allowed error. The VM's own incidental `log!`/`eprintln!`/`println!` traces are gone (undefined opcode and unknown loop state are now bails).
- `new_string_with` returns NIL and records an allocation error when an allocation budget is active (a raw `&mut String` cannot be capped before it reallocates); budgeted hosts use `new_bounded_string_with` / `temp_bounded_string_with`. Unbudgeted VMs are unchanged.

Test evidence (macOS, on 4146fc5ab):
- `cargo test -p makepad-script`: all suites ok, no FAILED/panicked lines — lib 56, new suites `equality_fuel_bail` 13, `try_catch` 19, `execution_limits` 7, plus inline parser/tokenizer/vm/thread/handle regressions.
- `cargo test -p makepad-widgets`: 211 passed / 5 failed — the same 5 that fail on unmodified `work` (desktop_style, grid x2, widget_tree x2); no new failure.
- `cargo check -p makepad-platform -p makepad-widgets`: clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JgSir4cQyaakzPju4h6smw


ymote:
Rebased onto `dev` (5e9a69794) and retargeted from `work` to `dev`: the previous head sat on an unrelated history, which is why the PR showed 5,693 files. It is the same seven commits, applied cleanly with no conflicts.

Re-run on this base (macOS):
- `cargo test -p makepad-script`: all suites ok — lib 56, `equality_fuel_bail` 13, `try_catch` 19, `execution_limits` 7, plus the inline parser/tokenizer/vm/thread/handle regressions.
- `cargo test -p makepad-widgets --lib`: 229 passed / 2 failed — the same two `grid` tests that fail on unmodified `dev` (228 / 2); no new failure.
- `cargo check -p makepad-platform -p makepad-widgets`: clean.

